PR #56, opened 2015-01-24
Add GitLab example to README

giedriusr:
Thought it would be useful to add an example in README about GitLab as well.


giedriusr:
Ups, haven't seen there was already one. (facepalm)

mmozuras:
@giedriusr no problem, it's the thought that matters :smile:

giedriusr:
Yeah, I know :-)

PR #58, opened 2015-01-29
Gitlab integration: comments are duplicated because of default 20-item pagination

siebertm:
I fixed a small issue which leads to duplicate comments on when using the GitlabFormatter: When you have more than 20 comments on a commit, pagination kicks in and not all comments are seen, thus duplicate checking breaks.

I just set per_page to 500 to quickly work around that issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Changes
+
+* [#58](https://github.com/mmozuras/pronto/pull/58): GitlabFormatter uses a high +per_page+ value to avoid pagination (and thus duplicate comments)
 
 ## 0.4.0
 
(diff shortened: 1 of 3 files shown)


mmozuras:
@jeroenj could you take a look?

jeroenj:
Nice find. 500 seems like a sensible default. I'd accept it. :)

siebertm:
Just a random number from the top of my head ;-)

mmozuras:
@siebertm could you add an entry to CHANGELOG?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Changes
```
mmozuras: I'd put it under `Bugs fixed`.

mmozuras:
Released with this fix as `0.4.1`.

PR #60, opened 2015-01-29
Sort runners alphabetically

andyw8:


mmozuras:
@andyw8 that makes sense :+1: :smile:

PR #70, opened 2015-02-18
Re-order methods of running in README

andyw8:
I think it makes sense to describe how to run Pronto locally first, since running against GitHub/GitLab requires more configuration. Thoughts?


mmozuras:
@andyw8 makes sense :+1:

PR #75, opened 2015-03-12
Possible fix for issue with Github PR formatter commenting on wrong lines

### LINKED ISSUE #39: Check pull request for deviation before writing comments
GithubPullRequestFormatter bug:

If new commits are added to a pull request, Pronto sometimes writes comments on incorrect lines. Before doing that, it should check whether pull request hasn't changed.

Reported by @ljank

11mdlow:
Alright, this is to follow up to my previous pr: https://github.com/mmozuras/pronto/pull/72 and is an attempt to fix #39.

The issue is still the same: pronto has issues where you make commits/changes around the same sections of code, as it will actually pick up a line from it's last context, and not when it was added. This then calculates an incorrect position, and comments on a wrong line.

I'll admit this isn't be ideal behavior, as it will now comment on an outdated diff, but to me it is better than commenting on the wrong line.

What do you think?

### DIFF
--- a/lib/pronto/formatter/github_pull_request_formatter.rb
+++ b/lib/pronto/formatter/github_pull_request_formatter.rb
@@ -14,7 +14,7 @@ module Pronto
           line = nil
           sha = commits.find do |commit|
             patches = repo.show_commit(commit)
-            line = patches.find_line(message.full_path, message.line.new_lineno)
+            line = patches.find_added_line_with_content(message.full_path, message.line.content)
             line
           end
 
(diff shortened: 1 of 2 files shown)


mmozuras:
@11mdlow sorry for taking so long to answer. Doing it the proposed way wouldn't work when there are multiple lines that are exactly same, right?

mmozuras:
PRs https://github.com/mmozuras/pronto/pull/91 and https://github.com/mmozuras/pronto/pull/93 from @jhass seems to be better fixes for this issue.

@11mdlow thanks for the effort! :bow:

PR #77, opened 2015-03-24
Support enterprise/non-standard Github URLs

nickcampbell18:
My company uses Github Enterprise, which is found on a URL like so: `https://github.mycompany.com`. 

This PR supports the passing of custom Github endpoints via environment variables (the variables mirror the names used in the [Octokit](https://github.com/octokit/octokit) library):

* `GITHUB_WEB_ENDPOINT` (default: https://github.com)
* `GITHUB_API_ENDPOINT` (default: https://api.github.com)

This commit also modifies the git remotes parser to look within custom hostnames instead of public github remotes.

### DIFF
--- a/lib/pronto/github.rb
+++ b/lib/pronto/github.rb
@@ -37,14 +37,16 @@ module Pronto
     def slug
       @slug ||= begin
         @repo.remote_urls.map do |url|
-          match = /.*github.com(:|\/)(?<slug>.*).git/.match(url)
+          match = /#{Regexp.escape(github_hostname)}(:|\/)(?<slug>.*).git/.match(url)
           match[:slug] if match
         end.compact.first
       end
     end
 
     def client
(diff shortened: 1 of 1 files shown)


nickcampbell18:
This PR is as much about starting a discussion. This scratched our immediate itch; I'd love to see it in the main gem, but I think at this point the code may be asking for a separate `Configuration` class.

mmozuras:
@nickcampbell18 sorry for taking a long while to answer. The need for Configuration (`.pronto.yml`) is clear. I'll try to tackle it for the next release. It shouldn't be very hard, but in case I'll fail, keep this PR and we'll stick with using `ENV` for now :smile:

Knack:
+1

mmozuras:
@nickcampbell18 I've added ability to configure Pronto via .pronto.yml (unreleased) and at the same time, solved this: #104

Thanks for this pull request, would not have happened without it! :bow:

PR #84, opened 2015-07-21
Require rugged 0.23.1

dankohn:
As a fix for https://github.com/libgit2/rugged/issues/475

It also requires this line for rugged in the Gemfile:
```ruby
  gem 'rugged', github: 'libgit2/rugged', ref: '233da19', submodules: true
```

### DIFF
--- a/pronto.gemspec
+++ b/pronto.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |s|
   s.require_paths = ['lib']
   s.executables << 'pronto'
 
-  s.add_runtime_dependency 'rugged', '~> 0.22.0'
+  s.add_runtime_dependency 'rugged', '~> 0.23.0'
   s.add_runtime_dependency 'thor', '~> 0.19.0'
   s.add_runtime_dependency 'octokit', '~> 3.8.0'
   s.add_runtime_dependency 'gitlab', '~> 3.4.0'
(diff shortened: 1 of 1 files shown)


mvz:
Looks like the build failed because rugged 0.23.0 wasn't out yet five days ago.

HaiTo:
i want it ;)

codebender:
:+1:

stim371:
I'm curious why you lock down the dependencies so tightly. `~> 0.22.0` stopped everyone from grabbing 0.23.0 when it came out.

If all of your gemspec dependencies only specify down to the minor version, (`~> 0.22`) it will allow users to upgrade all the way up to (but not including) the next major version of the gem, which should stop any breaking changes but allow installers to more fluidly specify and upgrade dependencies.

mvz:
> I'm curious why you lock down the dependencies so tightly.

I can understand this, since the major version is `0`. There's no real guarantee there that `0.23` will be compatible with `0.22`.

mmozuras:
Released as part of `0.4.3.`

Back from vacation, reviewing issues and continuing work on `0.5.0` :smile:

codebender:
@mmozuras Thanks! :smile_cat:

felixmohnert:
@mmozuras Yep, thanks a lot. I was now able to give pronto a try and it works perfectly :+1:

PR #86, opened 2015-08-26
Add env var for github slug

Zauberstuhl:
Pronto fetches the first url from all available remote urls.
If it is a fork it will be the upstream url

To support forks you should be able to configure this part of the url, too!

### DIFF
--- a/lib/pronto/github.rb
+++ b/lib/pronto/github.rb
@@ -35,7 +35,7 @@ module Pronto
     private
 
     def slug
-      @slug ||= begin
+      @slug = ENV['GITHUB_SLUG'] || begin
         @repo.remote_urls.map do |url|
           match = /.*github.com(:|\/)(?<slug>.*).git/.match(url)
           match[:slug] if match
--- a/lib/pronto/gitlab.rb
+++ b/lib/pronto/gitlab.rb
@@ -22,7 +22,7 @@ module Pronto
     private
 
     def slug
-      @slug ||= begin
+      @slug = ENV['GITLAB_SLUG'] || begin
         host = URI.split(endpoint)[2, 2].compact.join(':')
         slug = @repo.remote_urls.map do |url|
           match = /.*#{host}(:|\/)(?<slug>.*).git/.match(url)


tansaku:
:+1: for this!

tansaku:
@mmozuras any idea when we might be able to pull this in?  I'd switch our systems to using pronto, but this appears to be the key blocker for us ...

Zauberstuhl:
@tansaku you could use a temporarily solution like me -> https://github.com/diaspora/vines/blob/develop/Gemfile#L5

tansaku:
Good call @Zauberstuhl - I'm just checking that your code works for our situation - that would be a good plan if it does indeed work :-)

tansaku:
I've pulled down your fork, and am just trying to work out how to get an executable pronto out of your code ...

Zauberstuhl:
If you pulled the repository: `bin/pronto help`
If you installed it via bundler use: `bundle exec pronto help`

Check my travis config -> https://github.com/diaspora/vines/blob/develop/.travis.yml#L10

tansaku:
thanks - I get 

```
[tansaku@Samuels-MacBook-Pro:~/Documents/Github/pronto ((a747341...))]$ 
→ bin/pronto help
dyld: lazy symbol binding failed: Symbol not found: _SSL_load_error_strings
  Referenced from: /Users/tansaku/.rvm/gems/ruby-2.2.2/gems/rugged-0.22.2/lib/rugged/rugged.bundle
  Expected in: flat namespace

dyld: Symbol not found: _SSL_load_error_strings
  Referenced from: /Users/tansaku/.rvm/gems/ruby-2.2.2/gems/rugged-0.22.2/lib/rugged/rugged.bundle
  Expected in: flat namespace

Trace/BPT trap: 5
````

I'm on OSX. I was trying to run locally before starting to play with CI like travis - we have travis set up on the repos we're tracking, but was trying to slice off a simpler check first, since we already have hound running there ...

**Comment by Zauberstuhl** [comment]
@tansaku seams like an OSX problem checkout https://github.com/libgit2/rugged/issues/475 and or update rugged-gem to the newest version using bundler

I would suggest installing pronto via bundler, cause it installs automatically required dependencies.
Could save you some trouble :)

**Comment by tansaku** [comment]
thanks @Zauberstuhl - yeah, found that rugged thing too - just spent the last 10 trying alternatives - seems to be blocked for me on OSX at the moment whatever I do - newest rugged gem seems incompatible:

```
[tansaku@Samuels-MacBook-Pro:~/Documents/Github/pronto ((a747341...))]$ 
→ irb
Unable to activate pronto-0.4.2, because rugged-0.23.2 conflicts with rugged (~> 0.22.0), octokit-4.0.1 conflicts with octokit (~> 3.8.0)
2.2.2 :001 > require 'pronto'
Gem::ConflictError: Unable to activate pronto-0.4.2, because rugged-0.23.2 conflicts with rugged (~> 0.22.0), octokit-4.0.1 conflicts with octokit (~> 3.8.0)
```
I was assuming running `pronto` at the CLI will always hit my global gem, but I guess I can add `gem 'pronto', :git => 'https://github.com/Zauberstuhl/pronto.git'` to my local gemfile and then run `bundle exec pronto ...` giving it a go ...

Zauberstuhl:
> 'pronto', :git => 'https://github.com/Zauberstuhl/pronto.git' to my local gemfile and then run bundle exec pronto ... giving it a go ...

yes that should do the trick

tansaku:
working - okay, thanks for your help - great proof of concept for my team - now we just need @mmozuras to pull this in - for ref I can now post across forked repos like so:

GITHUB_SLUG=makersacademy/takeaway-challenge PULL_REQUEST_ID=124 bundle exec pronto run -f github_pr -c origin/master

Zauberstuhl:
don't forget the `GITHUB_ACCESS_TOKEN` but that should be it

tansaku:
I didn't seem to need the GITHUB_ACCESS_TOKEN as our repos are all public ...

mmozuras:
@Zauberstuhl would using `pull[:repo][:git_url]` (if available, check `pull_id` code) solve this issue too?

tansaku:
@Zauberstuhl what do you think of @mmozuras's suggestion above?

mmozuras:
@tansaku I've added ability to configure Pronto via `.pronto.yml` and at the same time, solved this: https://github.com/mmozuras/pronto/pull/104

Thanks @Zauberstuhl for this pull request, would not have happened without! :bow:

Zauberstuhl:
@mmozuras thanks and sorry haven't seen your post above

PR #87, opened 2015-08-26
handle github remote urls without .git suffix

modosc:
```Github#slug``` doesn't correctly parse remote urls if they don't end in ```.git``` even though cloning from github (at least, haven't tested any other servers) works happily with the same urls. if this is a universal thing i'm happy to make the same change for gitlab as well.


mmozuras:
@modosc could you add an entry to CHANGELOG?

mmozuras:
@modosc update `CHANGELOG` myself, no worries :smile: 

Thanks :bow:

PR #91, opened 2015-09-06
Pull request formatter: Find position in full diff and fix commit_id

jhass:
https://developer.github.com/v3/pulls/comments/#create-a-comment
appears to be misleading or wrong, depending on how you interpret it.

commit_id is meant to reference the "state" of the PR when creating the
commit, that is it references the most recent commit of the PR, always.

position is meant to be the position in the full diff, not the position
in any single commit's diff.

With full diff the combined diff of all commits is meant, as you get for
example via /:owner/:repo/pull/:id.diff

### DIFF
--- a/lib/pronto.rb
+++ b/lib/pronto.rb
@@ -34,7 +34,7 @@ module Pronto
 
     result = run_all_runners(patches)
 
-    formatted = formatter.format(result, repo)
+    formatted = formatter.format(result, repo, patches)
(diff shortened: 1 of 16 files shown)


mmozuras:
@jhass thanks, I'll take more time with this PR as I want to test some scenarios myself :smile:

mmozuras:
@jhass thanks! :bow:

PR #92, opened 2015-09-06
Ignore failed pull request comments

jhass:
The diff output of the local git version and Github is not always consistent
especially in areas where file renames happened,  Github tends to recognize
these better, leading to messages we can't post because their diff position is
non-existent on Github.

### DIFF
--- a/lib/pronto/formatter/github_pull_request_formatter.rb
+++ b/lib/pronto/formatter/github_pull_request_formatter.rb
@@ -24,6 +24,13 @@ module Pronto
         comments = client.pull_comments(sha)
         existing = comments.any? { |c| comment == c }
         client.create_pull_comment(comment) unless existing
+      rescue Octokit::UnprocessableEntity => e
+        # The diff output of the local git version and Github is not always
+        # consistent, especially in areas where file renames happened, Github
+        # tends to recognize these better, leading to messages we can't post
+        # because their diff position is non-existent on Github.
+        # Ignore such occasions and continue posting other messages.
+        STDERR.puts "Failed to post: #{comment.inspect} with #{e.message}"
       end
     end
   end


mmozuras:
@jhass I assume that GitHub is also using `libgit2`. We must be able to achieve the same result. Maybe passing the `patience` option for `diff` would make it consistent? Wanna try that?

jhass:
Sure, trying to reduce those failure is good, but I view that as a separate change. Maybe there's a logger I can throw a error or fatal add to not completely swallow it? What I want to get rid of is failure to post further valid comments if there's an invalid one in the middle.

jhass:
I added printing a warning to stderr.

mmozuras:
@jhass thanks, you'll be certainly be the MVP of Pronto `0.5.0`! :smile:

jhass:
Well, I wanted to move away from HoundCI since it ignores our configs all the time, so we tried running Pronto in an automated fashion and found these issues ;)

jhass:
Mmh, looking into preparing to try with the `patience` diff algo for when I run into a situation where a comment would fail again, rugged mentions the option but says it's not implemented? http://www.rubydoc.info/gems/rugged/Rugged%2FTree.diff

jhass:
But it does set it https://github.com/libgit2/rugged/blob/e5de8acfbaf159c043444804260f9ade753267a4/ext/rugged/rugged_diff.c#L130-L132, weird, we'll see I guess

mmozuras:
> Well, I wanted to move away from HoundCI since it ignores our configs all the time, so we tried running Pronto in an automated fashion and found these issues ;)

@jhass are those diaspora configs? :smile: 

> Mmh, looking into preparing to try with the patience diff algo for when I run into a situation where a comment would fail again, rugged mentions the option but says it's not implemented?

Seems to be implemented in libgit2, maybe only the docs are out of date: https://github.com/libgit2/libgit2/search?utf8=%E2%9C%93&q=patience

jhass:
>  are those diaspora configs?

Yeah ;)

mmozuras:
@jhass wow. Do tell if you'll end up switching to `Pronto` :smile:

jhass:
Well, Hound is disabled and we're running Pronto (with my patches) for a while now, check the activity of @diaspora-code-review ;)

PR #93, opened 2015-09-06
Fallback to original_position when pulling pull request comments

jhass:
position gets null if it's outdated


mmozuras:
Perfect, thanks! :+1:

PR #94, opened 2015-09-06
Auto paginate responses from the Github API

jhass:
By default Octokit only fetches the first 30 comments of a pull request,
on large pull requests that makes the duplicate comment detection fail,
leading to duplicates.


mmozuras:
:clap: thanks!

PR #101, opened 2015-10-14
Make gitlab work with ssh port urls `ssh://user@url.com:123/path/to/repo`

doits:
Just a simple change with spec to make it work with ssh port urls.


doits:
looks like travis got an error and it failed, but not because of a spec

mmozuras:
@doits could you rebase against master?

doits:
done

mmozuras:
@doits thanks! :bow:

PR #115, opened 2015-12-03
Add pronto-jscs to runner list

spajus:
@mmozuras


mmozuras:
@spajus amazing :smile:

PR #116, opened 2015-12-05
Improve github formatter error output

spajus:
`.inspect` wasn't a nice thing to see. Also added a test for it.


mmozuras:
@spajus thanks! :bow:

PR #123, opened 2016-01-07
Add pronto-swiftlint to README

ajanauskas:
@mmozuras this is second pronto runner for `Swift` language lint tool

Right now both `tailor` and `SwiftLint` seem pretty young projects. `SwiftLint` is more active, however both of them lack features compared to other languages like ruby (`rubocop`)

Will try to support both of the runners, maybe in the future it will be more clear which tool will be used more by `Swift` community


mmozuras:
@ajanauskas great, thanks! :bow:

PR #124, opened 2016-01-11
Update readme with working exclude example

spajus:
@mmozuras


mmozuras:
@spajus yeah, that's a way better example :smile:

PR #126, opened 2016-02-04
Add runner attribute to the message initialization

### LINKED ISSUE #113: Provide the adequate runner class in a message
Hi!

I need to have the runner class name inside each report message for a personal project. The simplest solution is to add an extra runner argument to the constructor of the `Message` class, but this requires to adjust all the existing runners. An other (very ugly) solution is to somehow [parse the backtrace](https://github.com/kami-zh/caller_class) in the `Message`'s constructor and get the caller class from it. 

Which solution sounds better for you? Would you accept a PR?

skateman:
Resolves #113

### DIFF
--- a/lib/pronto/message.rb
+++ b/lib/pronto/message.rb
@@ -1,10 +1,10 @@
 module Pronto
   class Message
-    attr_reader :path, :line, :level, :msg, :commit_sha
+    attr_reader :path, :line, :level, :msg, :commit_sha, :runner
 
     LEVELS = [:info, :warning, :error, :fatal]
 
-    def initialize(path, line, level, msg, commit_sha = nil)
+    def initialize(path, line, level, msg, commit_sha = nil, runner = nil)
       unless LEVELS.include?(level)
         raise ::ArgumentError, "level should be set to one of #{LEVELS}"
       end
@@ -13,6 +13,7 @@ module Pronto
       @line = line
       @level = level
       @msg = msg
+      @runner = runner
       @commit_sha = commit_sha
       @commit_sha ||= line.commit_sha if line
     end


spec/pronto/formatter/checkstyle_formatter_spec.rb
```diff
@@ -6,20 +6,34 @@ module Formatter
       describe '#format' do
         subject { formatter.format(messages, nil, nil) }
         let(:line) { double(new_lineno: 1, commit_sha: '123') }
-        let(:error) { Message.new('path/to', line, :error, 'Line Error') }
-        let(:warning) { Message.new('path/to', line, :warning, 'Line Warning') }
+        let(:error) do
+          Message.new(path:  'path/to',
+                      line:  line,
+                      level: :error,
+                      msg:   'Line Error')
```
mmozuras: Please use this style of alignment (consistent in project, easier to change, does not require editor plugins :smile:):

```ruby
Message.new(path: 'path/to',
            line: line,
            level: :error,
            msg: 'Line Error')
```

lib/pronto/message.rb
```diff
@@ -4,17 +4,19 @@ class Message
...
-      @line = line
-      @level = level
-      @msg = msg
-      @commit_sha = commit_sha
-      @commit_sha ||= line.commit_sha if line
+      @path = options[:path]
+      @line = options.fetch(:line, nil)
+      @level = options[:level]
+      @msg = options[:msg]
+      @runner = options.fetch(:runner, nil)
```
mmozuras: Using `options` instead of the way it was before, would require to change all the runners too. I'd prefer not have such a breaking change right now.

skateman: Yup, but then the number of arguments will trigger a rubocop warning, is that okay?

mmozuras: @skateman not perfect, but fine for now :smile:

skateman:
@mmozuras all fixed, now it's a quite small patch :)

mmozuras:
@skateman thanks! :smile:

PR #134, opened 2016-03-22
Colorize text formatter

### LINKED ISSUE #129: Text Formatter with Color
Currently the text formatter doesn't output in color but it would look much nicer if it did.

Given the standard (--colored) config, would that be an acceptable feature?

The way [Rubocop does it](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/formatter/simple_text_formatter.rb) using [Rainbow](https://github.com/sickill/rainbow) seems like a good pattern to follow, it wouldn't be that complex.

nbekirov:
This PR colorizes the text formatter to improve terminal readability when used locally.

Before
before

After
after

### Configuration

The new feature is controlled via `text_formatter`:`colors` property in `.pronto.yml`. This is also reflected in the `README.md`. By default it's disabled.

This feature is always disabled (configuration is ignored) when not in TTY to enable file I/O redirection, `grep`, etc.

N.B. I've had some hard time trying to enable it by default as there seems to be some issues with overriding truthy `ConfigFile::EMPTY` values via `.pronto.yml`/`ENV` settings. I'll investigate further and report back.

P.S. I've left the proposed `--colored` option out as it would require a bit of a bigger refactoring to centralize the configuration and its modification. Currently `TEXT_FORMATTER_COLORS=true pronto run` will be available as an alternative.

### New dependencies

I've used Rainbow as suggested in #129.

### Colorizable mixin

This is a reusable module intended to be used by any formatter that may need colorization. Its implementation is inspired by the one seen in Rubocop.

### Colors and styles

The colors I've used are inspired by Rubocop as I find them to be quite nice and recognizable.

Resolves #129

### DIFF
--- a/lib/pronto/formatter/colorizable.rb
+++ b/lib/pronto/formatter/colorizable.rb
@@ -1,0 +1,17 @@
+module Pronto
+  module Formatter
+    module Colorizable
+      def colorize(string, color)
+        rainbow.wrap(string).color(color)
+      end
+
+      private
+
+      def rainbow
+        @rainbow ||= Rainbow.new.tap do |rainbow|
+          rainbow.enabled = $stdout.tty?
+        end
+      end
+    end
+  end
+end
--- a/lib/pronto/formatter/text_formatter.rb
+++ b/lib/pronto/formatter/text_formatter.rb
@@ -1,22 +1,44 @@
 module Pronto
   module Formatter
     class TextFormatter
+      include Colorizable
+
+      LOCATION_COLOR = :cyan
+
+      LEVEL_COLORS = {
+        info: :yellow,
+        warning: :magenta,
+        error: :red,
+        fatal: :red
+      }.freeze
+
       def format(messages, _, _)
         messages.map do |message|
-          level = message.level[0].upcase
-          "#{location(message)} #{level}: #{message.msg}"
+          "#{format_location(message)} #{format_level(message)}: #{message.msg}".strip
         end
       end
 
       private
 
-      def location(message)
+      def format_location(message)
         line = message.line
         lineno = line.new_lineno if line
         path = message.path
-        commit_sha = message.commit_sha[0..6] if message.commit_sha
+        commit_sha = message.commit_sha
 
-        (path.nil? && lineno.nil?) ? commit_sha : "#{path}:#{lineno}"
+        if path || lineno
+          path = colorize(path, LOCATION_COLOR) if path
+          "#{path}:#{lineno}"
+        elsif commit_sha
+          colorize(commit_sha[0..6], LOCATION_COLOR)
+        end
+      end
+
+      def format_level(message)
+        level = message.level
+        color = LEVEL_COLORS.fetch(level)
+
+        colorize(level[0].upcase, color)
       end
     end
   end
--- a/lib/pronto.rb
+++ b/lib/pronto.rb
@@ -3,6 +3,7 @@ require 'octokit'
 require 'gitlab'
 require 'forwardable'
 require 'httparty'
+require 'rainbow'
 
 require 'pronto/gem_names'
 
@@ -25,6 +26,7 @@ require 'pronto/github'
 require 'pronto/gitlab'
 require 'pronto/bitbucket'
 
+require 'pronto/formatter/colorizable'
 require 'pronto/formatter/text_formatter'
 require 'pronto/formatter/json_formatter'
 require 'pronto/formatter/github_formatter'
--- a/pronto.gemspec
+++ b/pronto.gemspec
@@ -45,6 +45,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('octokit', '~> 4.3', '>= 4.1.0')
   s.add_runtime_dependency('gitlab', '~> 3.6', '>= 3.4.0')
   s.add_runtime_dependency('httparty', '~> 0.13.7')
+  s.add_runtime_dependency('rainbow', '~> 2.1')
   s.add_development_dependency('rake', '~> 11.0')
   s.add_development_dependency('rspec', '~> 3.4')
   s.add_development_dependency('rspec-its', '~> 1.2')
--- a/spec/pronto/formatter/colorizable_spec.rb
+++ b/spec/pronto/formatter/colorizable_spec.rb
@@ -1,0 +1,37 @@
+module Pronto
+  module Formatter
+    describe Colorizable do
+      let(:formatter_class) do
+        klass = described_class
+
+        Class.new(TextFormatter) do
+          include klass
+        end
+      end
+
+      let(:formatter) do
+        formatter_class.new
+      end
+
+      describe '#colorize' do
+        subject { formatter.colorize('Warning', :yellow) }
+
+        context 'in TTY' do
+          before { $stdout.stub(:tty?) { true } }
+
+          it 'colorizes the passed string' do
+            should eq("\e[33mWarning\e[0m")
+          end
+        end
+
+        context 'not in TTY' do
+          before { $stdout.stub(:tty?) { false } }
+
+          it 'returns the passed string' do
+            should eq('Warning')
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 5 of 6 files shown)


lib/pronto/formatter/colorizable.rb
```diff
@@ -0,0 +1,17 @@
...
+    module Colorizable
+      def colorize(string, color)
+        rainbow.wrap(string).color(color)
+      end
+
+      private
+
+      def rainbow
+        @rainbow ||= Rainbow.new.tap do |rainbow|
+          rainbow.enabled = config.text_formatter_colors && $stdout.tty?
```
tscheffe: Much more terse, and more clear, than the [equivalent rubocop cop code](https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/formatter/colorizable.rb#L10-L20). Nice job!

lib/pronto/formatter/text_formatter.rb
```diff
@@ -14,9 +30,30 @@ def location(message)
         line = message.line
         lineno = line.new_lineno if line
         path = message.path
-        commit_sha = message.commit_sha[0..6] if message.commit_sha
+        commit_sha = message.commit_sha
+
+        if path || lineno
```
tscheffe: You've introduced mixed levels of abstraction here.

`colorize(...)` is called in a method named `level` but also in methods named `format_location_*`. Logic around what to display is mixed between `location` and `format_location_*`.

I would suggest reducing it to only two layers, `format` which calls `format_level` and `format_location`. Inline `format_location_path` and `format_location_commit` into `format_location`, all they do now is add another layer of indirection.

Does that make sense? It's definitely not bad now, it would work as it. I just think it could have a little bit more unity in how things are laid out.

lib/pronto/formatter/text_formatter.rb
```diff
@@ -14,9 +30,30 @@ def location(message)
...
+          format_location_path(path, lineno)
+        elsif commit_sha
+          format_location_commit(commit_sha)
+        else
+          raise 'Invalid message location'
```
tscheffe: No exceptions were thrown previously, it would have just returned `nil` for location if the `commit_sha` wasn't present.

Is an exception something that should be introduced here?

I wouldn't really expect a "text formatter" to throw, I would expect it to do as much as possible and degrade gracefully.

If an exception should be thrown, is this the correct way to do it? It doesn't look like theres any exception handling code around the [call to format](https://github.com/mmozuras/pronto/blob/6961ced965b522fcee87f85ba0bd075fb2c4cde6/lib/pronto.rb#L51). That would indicate that the whole app is going to crash with an unhandled exception which isn't good. Maybe the `#format` method should catch exceptions and replace them with something like `"Unable to format message. Message: #{message} Error: #{e.inspect}"`.

lib/pronto/config.rb
```diff
@@ -4,7 +4,7 @@ def initialize(config_hash = ConfigFile.new.to_h)
       @config_hash = config_hash
     end
 
-    %w(github gitlab).each do |service|
+    %w(github gitlab text_formatter).each do |service|
```
tscheffe: `text_formatter` isn't a service, maybe that name should be updated to something like `nested_config`?

lib/pronto/formatter/text_formatter.rb
```diff
@@ -1,10 +1,26 @@
...
+      LEVEL_COLORS = {
+        info: :yellow,
+        warning: :magenta,
+        error: :red,
+        fatal: :red
+      }.freeze
+
+      attr_reader :config
+
+      def initialize(config = Config.new)
```
tscheffe: Where does this get called with a config being passed in?

lib/pronto/formatter/text_formatter.rb
nbekirov: This was something I was wondering about and decided to go with some short and specialized methods that also make rubocop & reek happier. 

Indeed you are correct that they also introduce layers/levels of abstractions and increase the cognitive load when comprehending such a simple code. I've addressed this in my latest commit.

lib/pronto/formatter/text_formatter.rb
nbekirov: > No exceptions were thrown previously, it would have just returned nil for location if the commit_sha wasn't present.

Correct! I've rolled back to this behavior (with updated spec).

lib/pronto/config.rb
nbekirov: Spot on!

I've replaced it with `configurable` because I just don't fine something like the following as nice:

```ruby
@config_hash[nested_config][key]
```

lib/pronto/formatter/text_formatter.rb
nbekirov: This is currently called here:
  * pronto/formatter/colorizable_spec.rb: [formatter_class.new(config)](https://github.com/mmozuras/pronto/pull/134/files#diff-0389d75965eaee0e29a946aa98b90797R15)
  * pronto/formatter/text_formatter_spec.rb: [described_class.new(config)](https://github.com/mmozuras/pronto/pull/134/files#diff-a8e1019a71e7c0c9aaccf75885afb994R5)

I've also tried to be consistent with other usages like:
  * pronto/lib/pronto/runners.rb: [def initialize(runners = Runner.runners, config = Config.new)](https://github.com/mmozuras/pronto/blob/master/lib/pronto/runners.rb#L3)

nbekirov:
@tscheffe Thank you for your thorough code review! Please check my replies and fixup commit. I hope you will like the end result better. :)

lib/pronto/formatter/text_formatter.rb
tscheffe: If this isn't constructed with a config outside of tests, how would I get a configured value other than the default of false?

I think theres a missing argument to the constructor somewhere that passes in the config, right?

lib/pronto/config.rb
tscheffe: `configurable`, way better! :+1:

lib/pronto/formatter/text_formatter.rb
tscheffe: Yeah, when I run your branch with `TEXT_FORMATTER_COLORS=true` it doesn't colorize the output for me :(

You've tested it and it's supposed too? Maybe I'm just missing something?

nbekirov: I'm following the flow described by http://guides.rubygems.org/ in [Make your own gem](http://guides.rubygems.org/make-your-own-gem/).

First build/create the gem:
```bash
pronto$ gem build pronto.gemspec && gem install pronto-0.6.0.gem;
  Successfully built RubyGem
  Name: pronto
  Version: 0.6.0
  File: pronto-0.6.0.gem
Successfully installed pronto-0.6.0
1 gem installed
```

Run it:
```bash
pronto$ ruby -Ilib ./bin/pronto run
```

Enable it via `.pronto.yml` with [valid YAML bool](http://yaml.org/type/bool.html)
```yml
text_formatter:
  colors: on
```

Enable it ENV variable (`true` or anything other):
```bash
pronto$ TEXT_FORMATTER_COLORS=true ruby -Ilib ./bin/pronto run
```

tscheffe: I got it working! Haven't built and run gems before, very cool.

tscheffe:
Everything is looking good to me! Works great!

Waiting on review from @mmozuras then right?

nbekirov:
Thanks, Trent!

@mmozuras 
I'm available to make any further changes needed. Also I can rebase(+squash) my commits upon your request.

mmozuras:
@nbekirov looks good 👍 
There's only thing I'm not sure about: it there a need for it to be configurable? In what kind of scenario, would I like to disable it manually?

nbekirov:
@mmozuras Indeed, the config toggle is not _needed_ per se - I'll remove it.

nbekirov:
@mmozuras I've removed the configuration, rebased, and squashed! Please let me know what do you think. :smiley:

mmozuras:
@nbekirov amazing, thanks 👏 😄

Issue #121, opened 2016-01-04
GitHub status formatter


dwradcliffe:
It would be really great if we could add custom statuses to a GitHub PR, one for each runner. This would make it much easier to see what failed at a glance, instead of lumping everything into one status.

I've started looking at the code to see what needs to change to make this happen.

This would depend on #113 and #83

dwradcliffe:
Here's an example:

![wip pronto by dwradcliffe](https://cloud.githubusercontent.com/assets/118850/12099473/bbb98164-b2f6-11e5-9fb6-871258cf1a6d.png)

mmozuras:
@dwradcliffe good idea!

Without making any changes, today this can be achieved by running Pronto multiple times, each time specifying a single runner. Not ideal, I know :smile:

dwradcliffe:
Yep, that was my plan until #83 is done. Still need #113 for my implementation to work.

mknapik:
What is the status of this enhancement?

mknapik:
@mmozuras @dwradcliffe I'm willing to submit PR if nobody has started working on this feature.

mmozuras:
@mknapik I haven't started working on it 😄

PR #142, opened 2016-04-13
Fix 'formattter' typo in spec names

mknapik:
While working on #121 I've noticed the spec names does not entirely correspond to the class names.

I hope this change won't cause any merge conflicts for other pending PRs.


mmozuras:
@mknapik wow, thanks 😄 🙇

PR #143, opened 2016-04-13
Add link to pronto-credo gem to README

carakan:
Another new member into this big family @mmozuras


mmozuras:
@carakan could you change it that it would be listed alphabetically? 😄

carakan:
I didn't realize it has a order, moved :smiley:

PR #144, opened 2016-04-14
Add GitHub status formatter

mknapik:
Solves #121.

The formatter creates one commit status per runner with information about issue counts.
Issues in status description are ordered by severity.

By default all Pronto Messages with level `:warning` or higher are considered a failed state for GitHub status (red status).
`GithubStatusFormatter` allows to override this behaviour by accepting `:level_mapping` option in the initializer (which is not supported when running the formatter directly with command line, e.g. `pronto run -f github_pr_status`).
There is no README info yet but I'll provide it once we agree on the interface and the way this formatter should work.

The result of running Pronto with the new formatter:
GithubStatusFormatter

@mmozuras Please let me know whether the PR adheres to the project conventions.
I got some RuboCop `line to long warnings` in specs, should I fix those?
Also I'm going to squash refactoring commits before merging.


lib/pronto/formatter/github_status_formatter.rb
```diff
@@ -0,0 +1,85 @@
+require_relative 'github_status_formatter/sentence'
+
+module Pronto
+  module Formatter
+    class GithubStatusFormatter
+      def initialize(opts = {})
+        @level_mapping ||= opts.fetch(:level_mapping, {})
```
mmozuras: Not sure if it's worth having a way to override it. Let's leave it off for now 😄

lib/pronto/formatter/github_status_formatter/sentence.rb
```diff
@@ -0,0 +1,42 @@
...
+        def to_s
+          case words.size
+          when 0
+            ''
+          when 1
+            words[0].to_s.dup
+          when 2
+            "#{words[0]}#{WORD_CONNECTORS[:two_words_connector]}#{words[1]}"
+          else
+            to_oxford_comma_sentence
```
mmozuras: Why have you chosen oxford comma?

mmozuras:
@mknapik amazing, thank you, honestly 👏 😄 

Just address a couple of comments, add info to README and CHANGELOG and it's gonna get merged! 😄

lib/pronto/formatter/github_status_formatter.rb
mknapik: The reason for implementing this was that Reek was returning only `warning`s even though stuff like `Doesn't depend on instance state (maybe move it to another class?) (UtilityFunction)` sound pretty subjective to me. Now when I think about it, it'd probably be better to address the issue at mmozuras/pronto-reek than here.

I'll get rid of `opts` parameter for now.

lib/pronto/formatter/github_status_formatter/sentence.rb
mknapik: I shamelessly ripped off and simplified the implementation from [ActiveSupport](http://apidock.com/rails/ActiveSupport/CoreExtensions/Array/Conversions/to_sentence).

mknapik:
@mmozuras Apart from adhering to your remarks I've also did some refactoring.
Additionally I renamed formatter cli name from `github_pr_status` to `github_status`.
Let me know what do you think.
I haven't squashed everything yet so you can review the changes easily.

README.md
```diff
@@ -103,13 +103,21 @@ or, if you want comments to appear on pull request diff, instead of commit:
 $ GITHUB_ACCESS_TOKEN=token PULL_REQUEST_ID=id pronto run -f github_pr -c origin/master
 ```
 
+Use `GithubStatusFormatter` to submit [commit status](https://github.com/blog/1227-commit-status-api):
+
+```sh
+$ GITHUB_ACCESS_TOKEN=token pronto run -f github_status -c origin/master
```
mmozuras: I think that it's worth showing that multiple formatters can be used here.

mmozuras:
@mknapik looks good, just change that one last thing 😄

aergonaut:
@mknapik @mmozuras would it be too much to ask to have the status contexts formatted more like how other services format them? i.e. instead of "Pronto::Rubocop", something like "pronto/rubocop" or "lint/pronto/rubocop" to be more similar to how Travis CI uses "continuous-integration/travis-ci/pr"

mknapik:
@aergonaut Actually I had no idea what was the convention.
The [docs](https://developer.github.com/v3/repos/statuses/) use `continuous-integration/jenkins` and `security/brakeman` as examples.
I guess I can change it to `pronto/$runner`.

README.md
mknapik: Ok, I'll add one more example.
Although I believe using multiple formatters is not documented anywhere else.

mknapik:
@mmozuras @aergonaut Regarding context names:
Do you think a simple function like `runner.class.downcase.gsub('::', '/')` is enough or should I implement something like [`ActiveSupport::Inflector#underscore`](http://apidock.com/rails/v4.2.1/ActiveSupport/Inflector/underscore)?

The difference:

Pronto Runner | downcase | underscore
--- | --- | ---
`Pronto::Brakeman` | `pronto/brakeman` | `pronto/brakeman`
`Pronto::RailsBestPractices` | `pronto/railsbestpractices` | `pronto/rails_best_practices`

aergonaut:
@mknapik looks good to me 😄

mmozuras:
@mknapik looks good, amazing contribution 👍 🙇

mknapik:
@mmozuras Thanks!
The gem is well written and contributing was a pleasure!

kevinjalbert:
This is awesome. Was trying to make it work using version 0.6.0, then realized that this isn't released yet 😉 .

I look forward to using it in the next version bump.

PR #150, opened 2016-04-22
Add runner name to the text formatter

nbekirov:
This PR adds runner name to the text formatter output.

### Motivation

When I'm reviewing my work using pronto with multiple runners enabled I often find myself fixing issues grouped not (only) by filename but by some kind of category - style, security, complexity, duplication, etc. I need a way to analyze, filter and order the output to suit my needs (using `grep`, `sort`, etc.). As the runner is currently not shown there is a gap in the message's context - I know the location, severity, etc. but not the reporter. Working around this currently involve starting `pronto` with the `--runner` option.
### Use case scenario

I want to run pronto on my branch once and fix `rubocop` issues first, then `reek` ones and completely ignore `jshint` at this stage. Additionally counting the messages from each runner would be nice. Also often there is a message that I don't agree with (and want to disable) or I don't understand (and need to check the docs) and I have no idea which runner reported it.

### DIFF
--- a/lib/pronto/formatter/text_formatter.rb
+++ b/lib/pronto/formatter/text_formatter.rb
@@ -14,7 +14,9 @@ module Pronto
 
       def format(messages, _, _)
         messages.map do |message|
-          "#{format_location(message)} #{format_level(message)}: #{message.msg}".strip
+          "#{format_location(message)} "\
+          "#{format_runner(message)}#{format_level(message)}: "\
+          "#{message.msg}".strip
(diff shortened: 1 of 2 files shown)


ibrahima:
Would also love to see this on other formatters; sometimes you have different policies for different linters (eg. Rubocop violations are not allowed, Reek/RailsBP violations are sometimes okay) so it's nice to know which linter a comment came from so you know whether you need to address it or not.

PR #152, opened 2016-04-25
Replace PNG badges with SVG in README

d-unseductable:
- Use Retina-ready icons for badges

... because why not 😄


mmozuras:
@d-unseductable exactly, why not! 🙇 Thanks 😄

PR #153, opened 2016-05-01
Fix Popper Issue mmozuras/pronto-poper#5

abinoam:
ruby_patches method sometimes returns an empty Array.
Calling #first on it will return nil that can cause errors.
So, using the bare @patches instance variable.

Fixes mmozuras/pronto-poper#5


mmozuras:
@abinoam 👍 thanks for a fix! 🙇

abinoam:
👍 Thanks for merging and for developing Pronto!

PR #157, opened 2016-06-01
Discover repository directory inside subdirectories

ibrahima:
Uses Rugged::Repository#discover to discover the location of the git
repository if the current working directory is a subdirectory of the git
repository. Working directory is changed to the root of the repository
so that runners pick up the appropriate configuration files.

This allows running pronto within subdirectories of a git repository.

### DIFF
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -49,7 +49,10 @@ module Pronto
 
       formatters = ::Pronto::Formatter.get(options[:formatters])
       commit = options[:index] ? :index : options[:commit]
-      messages = ::Pronto.run(commit, '.', formatters, path)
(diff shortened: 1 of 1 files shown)


doomspork:
@ibrahima how is the configurations handled for this?  I've got a project with subdirectories and I've had to duplicate Pronto configuration between some of the directories in order to get things working.

ibrahima:
Right, the point of this PR is to detect the repository root so that as long as your configuration files are in the root it'll always pick them up.

doomspork:
Woohoo!  @ibrahima, that's awesome.  👍

mmozuras:
@ibrahima awesome! 👏

ibrahima:
Thanks!

PR #154, opened 2016-05-10
Consolidate PR comments

doomspork:
Hi @mmozuras, here are the changes I made regarding #147.  I figured a PR was a good way to get and incorporate feedback.  You'll see I removed the checkboxes in favor of a bulleted list.

PR #159, opened 2016-06-07
Add more detail to GitHub messages

aergonaut:
This adds more detail to comments generated by the GitHub commit comment
and GitHub PR comment formatters by including the name of the Pronto
runner that generated the message and the warning level of the message.

In situations where a repo is using many different runners to report on
various different checks, it can be difficult to know which tool
generated which message. Adding the name of the runner allows this to be
more easily seen.

Adding the warning level to the message also could help to suggest which
messages are higher or lower priority if there are many messages being
generated.


mekhovov:
@mmozuras can you merge this PR please?

doomspork:
@aergonaut could you provide a screenshot of what the resulting comments look like?

aergonaut:
@doomspork There's an example of the generated text in the comments in the code

> Generated by Pronto::Rubocop. Severity: ⚠️ warning. Lorem ipsum dolor sit

doomspork:
@aergonaut I wanted to see what it would appear like in GitHub's UI rather than go by what the comments are showing.  

I'm curious how it will look when combined with #154, my initial concern is that it'd get _real_ noisy real quick with the extra meta data and a bunch of emojis.  Just my $0.02 though.

mknapik:
@aergonaut This is exactly what we're missing at the moment. Are you planning to continue working on that? (e.g. rebase)
@mmozuras Are you planning to merge it?

mknapik:
@aergonaut @mmozuras bump

davejlong:
This would be awesome to see in the GitLab formatter as well.

aergonaut:
Sorry to let this PR fall off, I'm still interested in fixing it back up and getting it into a merge-able state. I'm now thinking the format might be too verbose and maybe a more concise format that still shows the same amount of data might be better.

davejlong:
What led me to this issue is that I simply wanted which runner threw the error. Currently to fix and retest a Reek issue, I have to run Brakeman, Rubocop and Reek, locally which takes time.

doomspork:
@davejlong I've found googling the exact error message almost always leads me to the culprit.  

I do think this feature would be useful but I stand by my original comments with regards to the amount of noise this would produce.  If we could make the message configurable, that would suit most everyone's needs I think.

@aergonaut could this be used in the other formatters to maintain consistency?

aergonaut:
@doomspork More configurable messages sounds like a good idea, I agree. Should I try to look at making Pronto in general expose some message format configuration options? Maybe I could make `Comment` store more information, or delegate to the `Message` class, and maybe use a configurable format string for the body, for example `"%{runner} (%{level}): %{message}"`.

doomspork:
@aergonaut I like the idea of a configurable message with a formatted string like you've suggested!

mmozuras:
@aergonaut I like the idea of configurable format very much!

I would expect it to work with every formatter, solving [this PR](https://github.com/mmozuras/pronto/pull/150) too 😄.

aergonaut:
Cool sounds great. I'll close this PR and open another one with a more generic solution!

PR #162, opened 2016-06-20
don't use an error exit code if all the messages are info.

dnl:
Hi, 

I'm having this issue with exit code on info messages.

The doc implies that using the `--exit-code` flag, will only fail for errors and warnings.
```
--exit-code	Exits with non-zero code if there were any warnings/errors.
```

It was also returning non-zero exit code if all the messages are `info` level messages (in my case, pronto-rubocop 'refactor' severity)

Thanks for your time. 

Dan.

### DIFF
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -53,7 +53,10 @@ module Pronto
       messages = Dir.chdir(repo_workdir) do
         ::Pronto.run(commit, '.', formatters, path)
       end
-      exit(1) if options[:'exit-code'] && messages.any? { |m| m.level != :info }
+      if options[:'exit-code']
+        error_messages_count = messages.count { |m| m.level != :info }
+        exit(error_messages_count)
+      end
     rescue Rugged::RepositoryError
       puts '"pronto" should be run from a git repository'
     end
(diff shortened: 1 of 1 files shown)


lib/pronto/cli.rb
```diff
@@ -53,7 +53,7 @@ def run(path = nil)
       messages = Dir.chdir(repo_workdir) do
         ::Pronto.run(commit, '.', formatters, path)
       end
-      exit(messages.count) if options[:'exit-code']
+      exit(1) if options[:'exit-code'] && messages.any? { |m| m.level != :info }
```
mmozuras: Why did you change `messages.count` into `1`? 😄 
Let's tackle this separately.

mmozuras: @doomspork could you give a source that would back up this statement? 😄 

From what I understand, there's only one hard rule: `0` means `success`. Other exit codes can be defined by each application individually. Reference: http://stackoverflow.com/a/1101969

dnl:
I've updated this to output the exit code as the number of errors

mmozuras:
@dnl thanks! 🙇 👍

PR #173, opened 2016-08-13
add pronto-eslint_npm and a hint about the difference from pronto-eslint

doits:


mmozuras:
@doits perfect, thanks! 😄

Issue #170, opened 2016-08-07
422 - Validation Failed When using GithubPullRequestFormatter


samnang:
Here is the output when I run:

```
Failed to post: POST https://api.github.com/repos/user_a/repo1/pulls/49/comments: 422 - Validation Failed
Error summary:
  resource: PullRequestReviewComment
  code: invalid
  field: position // See: https://developer.github.com/v3/pulls/comments/#create-a-comment
66 Pronto messages posted to GitHub
Failed to post: POST https://api.github.com/repos/user_a/repo1/pulls/49/comments: 422 - Validation Failed
Error summary:
  resource: PullRequestReviewComment
  code: invalid
  field: position // See: https://developer.github.com/v3/pulls/comments/#create-a-comment
66 Pronto messages posted to GitHub
```

When I look at Github API documentation, it seems `position` is required, but it seem it's missing when creating comment here https://github.com/mmozuras/pronto/blob/809ed1d50be1632fa2f3885eb13d1ff865158a0e/lib/pronto/formatter/git_formatter.rb#L66-L72

jglapa:
Getting the exact same error!

jglapa:
looks like duplicate to #149

mmozuras:
Fixed and released [v0.7.1](https://github.com/mmozuras/pronto/releases/tag/v0.7.1).

samnang:
@mmozuras thanks for fixing it. Love it!

Issue #149, opened 2016-04-20
422 - Validation Failed


gingerlime:
Pronto looks awesome. I'd love to use it, but am hitting a small problem trying to push comments to github.

Here's my yml:

```yml
all:
  exclude:
    - 'spec/**/*'
github:
  slug: gingerlime/my-private-repo
  # access_token: B26354
  api_endpoint: https://api.github.com/
  web_endpoint: https://github.com/
max_warnings: 150
verbose: true
```

I'm running

```
GITHUB_ACCESS_TOKEN=*** pronto run -f github_pr -c origin/master
```

And it's showing these errors (across several files):

```
Running Pronto::Rubocop
Creating pull request comment on 1
Failed to post: [17a30b6181c4678efccdcbd1514b052560ab9c3e] app/controllers/articles_controller.rb:21 - Align the elements of a hash literal if they span more than one line. with POST https://ap
i.github.com/repos/gingerlime/my-private-repo/pulls/1/comments: 422 - Validation Failed
Error summary:
  resource: PullRequestReviewComment
  code: invalid
  field: path // See: https://developer.github.com/v3/pulls/comments/#create-a-comment
Creating pull request comment on 1
Failed to post: [17a30b6181c4678efccdcbd1514b052560ab9c3e] app/controllers/articles_controller.rb:88 - Avoid using `rescue` in its modifier form. with POST https://api.github.com/repos/gingerli
me/my-private-repo/pulls/1/comments: 422 - Validation Failed
Error summary:
  resource: PullRequestReviewComment
  code: invalid
  field: path // See: https://developer.github.com/v3/pulls/comments/#create-a-comment
Creating pull request comment on 1
Failed to post: [17a30b6181c4678efccdcbd1514b052560ab9c3e] app/controllers/articles_controller.rb:89 - Avoid using `rescue` in its modifier form. with POST https://api.github.com/repos/gingerli
me/my-private-repo/pulls/1/comments: 422 - Validation Failed
...
```

Any clue why it's returning this error? I think the token is set up correctly, but maybe I missed something.

steverob:
Same here. THere were couple of comments that were successfully created though.

steverob:
One thing I noticed was that this happened whenever pronto was trying to comment on lines that were not part of the diff of the PR in question.

Probably pronto is scanning the entire files that are under the changed list? @gingerlime

doomspork:
[...]

PR #174, opened 2016-08-15
Prevent crashing when files were removed

milgner:
When a file was removed in the changeset, pronto would try to open it, leading to errors like

    Errno::ENOENT: No such file or directory @ rb_sysopen - ...

This patch makes sure that the method returns `false` as expected.

### DIFF
--- a/lib/pronto/runner.rb
+++ b/lib/pronto/runner.rb
@@ -40,7 +40,7 @@ module Pronto
       return false if File.directory?(path)
       line = File.open(path, &:readline)
       line =~ /#!.*ruby/
-    rescue ArgumentError, EOFError
+    rescue ArgumentError, EOFError, Errno::ENOENT
       false
     end
   end


mmozuras:
@milgner could you expand on when it happens? 

Because ideally, it should never happen. Pronto only cares about what was added. If a file was removed, Pronto should not try to open it.

milgner:
Unfortunately I did not dig very deep into this. But one of our changesets removed a file and pronto tried to open it and crashed. The backtrace lead me to this location and after applying the patch, the error disappeared. I can try to find the output of our CI to illustrate but right now I did not find a way to quickly search all builds for a specific message.

mmozuras:
@milgner having a repro would be immensely helpful. I think that what you describe could potentially mask some greater underlying issue, which I would like to try and find.

milgner:
I totally understand! Right now I just came across a situation that might be related: Pronto is annotating problems on the removed part of the changeset, saying for example "Line is too long. [81/80]" on an old view template where the updated version isn't that long anymore. If it is looking at old file versions there then it might be the same code path that tried to open a deleted file, too.

mmozuras:
@milgner which version are you using?

milgner:
Current version in `Gemfile.lock` is 0.7.0.

mmozuras:
@milgner it could be another manifestation of an already fixed bug (#149, #170) in `0.7.0`. Could you update to `0.7.1` and check if it solves the issue?

mmozuras:
@milgner any news? 😄

milgner:
After updating to `0.7.1` the problems seem to have been tackled. Thanks for the assist, I'll close this PR.

PR #181, opened 2016-10-05
Support ENV for all config options

doomspork:
Howdy @mmozuras!  Following up on the conversation [here](https://github.com/mmozuras/pronto/pull/154#r75576452), this PR enables support for ENV variables on all configuration options.

### DIFF
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -12,7 +12,8 @@ module Pronto
     end
 
     def consolidate_comments?
-      !@config_hash['consolidate_comments'].nil?
+      consolidated = ENV['CONSOLIDATE_COMMENTS'] || @config_hash['consolidate_comments']
+      !(consolidated).nil?
     end
 
     def excluded_files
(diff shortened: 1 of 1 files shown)


lib/pronto/config.rb
```diff
@@ -12,6 +12,7 @@ def initialize(config_hash = ConfigFile.new.to_h)
     end
 
     def consolidate_comments?
+      @config_hash['consolidate_comments'] ||= ENV['CONSOLIDATE_COMMENTS']
```
doomspork: Should `ENV` variables take precedence?

mknapik:
How about prefixing each `ENV` variable with `PRONTO_` to avoid conflicts?

doomspork:
@mknapik the existing configuration variables don't rely on a prefix but I think that's a good idea, I'll leave that up to @mmozuras

lib/pronto/config.rb
mmozuras: @doomspork yes - they do everywhere else 😄

mmozuras:
@mknapik sounds like a good idea. Could you (or @doomspork) make a separate pull request for it?

doomspork:
@mmozuras I can update this PR.  Do you want me to prefix _all_ the configuration options (a breaking change)?

mmozuras:
@doomspork nah, just update according to my one comment. And then let's make a separate PR with _all_ options prefixed.

doomspork:
@mmozuras sounds good! 👍

doomspork:
Updated @mmozuras 👍

mmozuras:
@doomspork thanks! 🙇

PR #183, opened 2016-10-14
Add phpcs and phpmd runners

EllisV:


mmozuras:
@EllisV thanks! Amazing to see another language (PHP) added to the list 🙇

PR #185, opened 2016-10-17
Allow excluding files to lint for single runner

astrauka:
@mmozuras Allow excluding files to lint for single runner

```
eslint:
  exclude:
    - 'app/assets/**/*'
```


lib/pronto/version.rb
```diff
@@ -1,6 +1,6 @@
 module Pronto
   module Version
-    STRING = '0.7.1'.freeze
+    STRING = '0.7.2'.freeze
```
astrauka: eager to lift the version to `0.8.0` but this requires pronto-rubocop gem update :/

astrauka:
@mmozuras any feedback?

doomspork:
This functionality would be tremendously useful with how we use Pronto.

mmozuras:
@astrauka looks great, thanks! 🙇

doomspork:
Thank you @astrauka and @mmozuras!

PR #187, opened 2016-10-18
Allow for nil and false values for consolidation

doomspork:
As @mknapik pointed out [here](https://github.com/mmozuras/pronto/pull/154#r83859578) the existing config changes I made were not sufficiently checking for `nil` _and_ `false`, this PR changes that.

Thoughts @mmozuras?

### DIFF
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -12,8 +12,10 @@ module Pronto
     end
 
     def consolidate_comments?
-      consolidated = ENV['PRONTO_CONSOLIDATE_COMMENTS'] || @config_hash['consolidate_comments']
-      !(consolidated).nil?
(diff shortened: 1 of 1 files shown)


mknapik:
@doomspork Wow, that was fast :-)
I think: `@config_hash['consolidate_comments']` is the same as `@config_hash['consolidate_comments'] || false`.
If you like being explicit how about using `Hash#fetch`?: 
```ruby
@config_hash.fetch('consolidate_comments', false)
```

doomspork:
Good idea @mknapik, updated the code 👍

mmozuras:
@mknapik @doomspork 👍

doomspork:
Thanks again @mmozuras

PR #189, opened 2016-10-18
Fix nil consolidated comment

### LINKED ISSUE #188: GitHub Pull Request formatter fails if comments already exist
I got the following error on my CI:
```
/.rvm/gems/ruby-2.3.1/gems/pronto-0.7.1/lib/pronto/github.rb:35:in `create_pull_comment': undefined method `body' for nil:NilClass (NoMethodError)
```

After some debugging it turned out the error occurred for already existing comments.
Further troubleshooting showed it might have been related to `consolidate_comments` option.
I've created example project with 2 pull request:
* mknapik/pronto-github-nil-comment-issue#1 fails on second Pronto run
* mknapik/pronto-github-nil-comment-issue#2 works fine

The only difference is PR1 uses [comments consolidation](https://github.com/mknapik/pronto-github-nil-comment-issue/pull/1/commits/97873954574da44f0ea0f1dea8a02cc98d4d8eb5).
`GitFormatter#remove_duplicate_comments` returns `nil` for each commit which is already present.

CC: @doomspork

mknapik:
If all comments already existed,
consolidation feature transformed empty list into nil.

Fixes #188

### DIFF
--- a/lib/pronto/formatter/git_formatter.rb
+++ b/lib/pronto/formatter/git_formatter.rb
@@ -83,7 +83,7 @@ module Pronto
           existing = old_comments[key]
           comments = dedupe_comments(existing, comments) if existing
 
-          if config.consolidate_comments?
+          if config.consolidate_comments? && !comments.empty?
             comment = consolidate_comments(comments)
             memo.push(comment)
           else


mknapik:
Probably we should but I found it difficult to setup a spec for formatter.
I think we should consider extracting deduplication and consolidation logic to separate classes.

mknapik:
See my proposal for refactoring: https://github.com/mknapik/pronto/commit/199346a6334c0954f8d6bb3fa3885d63cfa8fd93
No specs yet.

doomspork:
@mknapik I like your refactor and the only feedback I have would be perhaps we can use something other than `#call` as the primary method, at first glance I thought you were using `Method.call`.

mmozuras:
@mknapik overall, I like the idea for the refactoring. I'd have a couple of nits to pick, but will save them for the PR 😄

mknapik:
At the moment I don't have time to finish the refactoring.
How about fixing the bug first and in the next PR I'll try to refactor.

mmozuras:
@mknapik 👌

mknapik:
@mmozuras Thanks!
When are you planning the next release?

mmozuras:
@mknapik this year 🙃. Hopefully, mid-November, but don't want to promise anything.

PR #190, opened 2016-10-20
Allow HTTParty 0.14

lautis:
HTTParty 0.14 removes dependency to JSON gem and is thus compatible with Ruby 2.4.

Gitlab 3.7 locks to HTTParty 0.13, but the next version will drop that requirement. Using Gitlab 3.6 should also work.


mmozuras:
@lautis 👍, thanks!

PR #198, opened 2016-12-23
Add ShellCheck runner to README

pclalv:


mmozuras:
@pclalv just a week ago I've talked with a former colleague about how great it would be to have a ShellCheck Pronto runner 😄.

Thanks! 🙇

PR #199, opened 2016-12-29
Test against ruby 2.4.0

andreas-venturini:


mmozuras:
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
An error occurred while installing json (1.8.3), and Bundler cannot continue.
Make sure that `gem install json -v '1.8.3'` succeeds before bundling.
```

We seem to depend on `gitlab` updating `httparty` dependency, which uses `json`.

mmozuras:
@andreas-venturini thanks! 🙇

PR #200, opened 2017-01-02
Change example configuration for gitlab/slug

stevejackson:
The example slug for gitlab is incorrect. The gitlab API looks up projects by its project id alone - the
username is never used.

It looks like there's been several issues reported about this already, so thought I'd make the docs a bit more clear.

I've enjoyed using pronto, thanks!


mmozuras:
@stevejackson thanks! 🙇

PR #204, opened 2017-02-22
Adds configurable option to set message format

aergonaut:
This allows the format of the messages produced by each formatter to be configured by specifying a format string in the .pronto.yml config file.

Format can be overridden on a global basis by specifying the `format` key at the top-level of the file, or it can be overridden on a per-formatter basis by nesting `format` keys under the formatter's name (as defined in `Formatter::FORMATTERS` constant). If a formatter does not have a specific format defined in the config file, it will fall back to the global format, and then fall back to the default format (which is just `"%{msg}"`).

Format uses a Ruby format string and makes all of the attributes of the `Message` object available for interpolation. The `TextFormatter` additionally adds a special `TextMessageDecorator` to add colorized and specially formatted values for file location and level. These were in the original formatter code so I wanted to keep them.

The Null, JSON, and Checkstyle formatters are excluded from this PR because those aren't really outputting a human-readable format, but rather transforming the message into some other format for some other system to consume. So specifying a format string for these didn't make sense to me.

### DIFF
--- a/lib/pronto/formatter/text_formatter.rb
+++ b/lib/pronto/formatter/text_formatter.rb
@@ -1,52 +1,13 @@
+require 'pronto/formatter/text_message_decorator'
+
 module Pronto
   module Formatter
     class TextFormatter < Base
-      class TextMessageDecorator < SimpleDelegator
-        include Colorizable
-
-        LOCATION_COLOR = :cyan
-
-        LEVEL_COLORS = {
-          info: :yellow,
-          warning: :magenta,
-          error: :red,
-          fatal: :red
-        }.freeze
-
-        def to_h
-          original = __getobj__.to_h
-          original[:color_level] = format_level(__getobj__)
-          original[:color_location] = format_location(__getobj__)
-          original
-        end
-
-        private
-
-        def format_location(message)
-          line = message.line
-          lineno = line.new_lineno if line
-          path = message.path
-          commit_sha = message.commit_sha
-
-          if path || lineno
-            path = colorize(path, LOCATION_COLOR) if path
-            "#{path}:#{lineno}"
-          elsif commit_sha
-            colorize(commit_sha[0..6], LOCATION_COLOR)
-          end
-        end
-
-        def format_level(message)
-          level = message.level
-          color = LEVEL_COLORS.fetch(level)
-
-          colorize(level[0].upcase, color)
-        end
-      end
-
       def format(messages, _, _)
         messages.map do |message|
-          (config.message_format(self.class.name) % TextMessageDecorator.new(message).to_h).strip
+          message_format = config.message_format(self.class.name)
+          message_data = TextMessageDecorator.new(message).to_h
+          (message_format % message_data).strip
         end
       end
     end
--- a/lib/pronto/formatter/text_formatter.rb
+++ b/lib/pronto/formatter/text_formatter.rb
@@ -1,44 +1,53 @@
 module Pronto
   module Formatter
-    class TextFormatter
-      include Colorizable
+    class TextFormatter < Base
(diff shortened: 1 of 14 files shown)


aergonaut:
This is the follow-up to https://github.com/mmozuras/pronto/pull/150 and https://github.com/mmozuras/pronto/pull/159

cc @nbekirov @doomspork @mmozuras

aergonaut:
Looks like Travis ran into a 404 error trying to post Pronto results in the after_success hook 😳

lib/pronto/formatter/text_formatter.rb
```diff
@@ -1,44 +1,53 @@
 module Pronto
   module Formatter
-    class TextFormatter
-      include Colorizable
+    class TextFormatter < Base
+      class TextMessageDecorator < SimpleDelegator
```
mmozuras: I'm usually not a fan of classes within classes. Could you move outside to `/formatter/text_message_decorator.rb` or give an argument why would you want to keep it inside the `TextFormatter` class?

lib/pronto/formatter/text_formatter.rb
```diff
@@ -1,44 +1,53 @@
...
       end
 
-      def format_level(message)
-        level = message.level
-        color = LEVEL_COLORS.fetch(level)
-
-        colorize(level[0].upcase, color)
+      def format(messages, _, _)
+        messages.map do |message|
+          (config.message_format(self.class.name) % TextMessageDecorator.new(message).to_h).strip
```
mmozuras: Could you split this line into multiple for readability? 😄

lib/pronto/formatter/text_formatter.rb
```diff
@@ -1,44 +1,53 @@
...
+        LEVEL_COLORS = {
+          info: :yellow,
+          warning: :magenta,
+          error: :red,
+          fatal: :red
+        }.freeze
+
+        def to_h
+          original = __getobj__.to_h
+          original[:color_level] = format_level(__getobj__)
```
mmozuras: Trying to understand pros and cons: why did you decide to leave `color_level` and `color_location` off README? 😄

mmozuras:
> Looks like Travis ran into a 404 error trying to post Pronto results in the after_success hook 😳

@aergonaut will look into it 👌

lib/pronto/formatter/text_formatter.rb
aergonaut: I nested it just because the decorator was only going to be used in this one place so I thought keeping it close improved readability. I don't feel strongly about it though so I'll reorganize it.

lib/pronto/formatter/text_formatter.rb
aergonaut: 👌

lib/pronto/formatter/text_formatter.rb
aergonaut: @mmozuras moved `TextMessageDecorator` to `pronto/formatter/text_message_decorator.rb`

lib/pronto/formatter/text_formatter.rb
aergonaut: Added to README 😃

lib/pronto/formatter/text_formatter.rb
aergonaut: Split up 💅

aergonaut:
@mmozuras I've updated the code according to your comments, please have a look!

mmozuras:
@aergonaut perfect, thank you! 🙇

jeroenj:
@aergonaut awesome addition! 👏 👍

PR #215, opened 2017-03-20
Do not mutate patches when excluding single runner files

astrauka:
Do not mutate patches when excluding single runner files

With mutation single runner excluded patches are removed from `@patches`
and not run for following runners.

@mmozuras @juno returning a new object instead of mutating existing one.

### DIFF
--- a/lib/pronto/git/patches.rb
+++ b/lib/pronto/git/patches.rb
@@ -15,8 +15,8 @@ module Pronto
         @patches.each(&block)
       end
 
-      def reject!(&block)
-        @patches.reject!(&block)
+      def reject(&block)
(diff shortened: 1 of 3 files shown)


lib/pronto/git/patches.rb
```diff
@@ -19,6 +19,10 @@ def reject!(&block)
         @patches.reject!(&block)
       end
 
+      def reject(&block)
```
mmozuras: @astrauka do we still need the `reject!` method in this class then?

mmozuras:
@astrauka makes sense, could you add a couple of simple specs for this?

astrauka:
@mmozuras updated

mmozuras:
@astrauka perfect, thanks! 🙇

PR #216, opened 2017-03-22
Add Bitbucket Server pull requests formatter

andrey-skat:


lib/pronto/clients/bitbucket_server_client.rb
```diff
@@ -0,0 +1,64 @@
...
+        line: position,
+        lineType: 'ADDED',
+        path: path,
+        srcPath: path
+      }
+    }
+    self.class.post(url, body: body.to_json, headers: @headers)
+  end
+
+  def paged_request(url, query = {})
```
mmozuras: @andrey-skat what's the reason for doing `paged_request`? Should we use it for `BitbucketClient` too?

lib/pronto/clients/bitbucket_server_client.rb
```diff
@@ -0,0 +1,64 @@
...
+    Enumerator.new do |y|
```
mmozuras: `y` is not a great variable name 😄 
Could you change it to something more descriptive?

lib/pronto/clients/bitbucket_server_client.rb
```diff
@@ -0,0 +1,64 @@
+class BitbucketServerClient
```
mmozuras: Could parts of this class be extracted to reduce duplication between `BitbucketServerClient` and `BitbucketClient`?

lib/pronto/bitbucket_server.rb
```diff
@@ -0,0 +1,34 @@
+module Pronto
+  class BitbucketServer < Bitbucket
+    def pull_comments(sha)
+      @comment_cache["#{pull_id}/#{sha}"] ||= begin
+        client.pull_comments(slug, pull_id).map do |comment|
+          if comment['commentAnchor']
```
mmozuras: What's a `commentAnchor`?

lib/pronto/bitbucket_server.rb
```diff
@@ -0,0 +1,34 @@
...
+            Comment.new(sha, comment['comment']['text'], comment['commentAnchor']['path'], comment['commentAnchor']['line'])
+          else
+            Comment.new(sha, comment['comment']['text'])
```
mmozuras: Why do we need to access properties via `['comment']['text']`?
What would it take to make it more similar to `bitbucket.rb`, accessing it like `comment.text`?

andrey-skat: Bitbucket Server can return not only inline comments, but comments for file and pull request itself, so we need to handle that comments.
I can rewrite this using local variable or use inject instead of map for exclude theirs

andrey-skat: Also format of returned data is different between bitbucket and bitbucket server

lib/pronto/bitbucket_server.rb
andrey-skat: It contains data about comment place (file path, line number, etc)

lib/pronto/clients/bitbucket_server_client.rb
andrey-skat: It's a `yielder`. I'll rename it

lib/pronto/clients/bitbucket_server_client.rb
andrey-skat: Bitbucket Server use paged API for get responses, so we must iterate over many pages

lib/pronto/clients/bitbucket_server_client.rb
andrey-skat: I don't see too much duplication between these two. They only have common interface

andrey-skat:
I'll try to add some specs :)

andrey-skat:
I've updated pull request. 
Not sure if tests are good :pensive:

mmozuras:
@andrey-skat thanks, they're fine! 😄

Issue #220, opened 2017-04-08
Pronto detects only some of the violations


sauliusgrigaitis:
Pronto detects only some of the violations. I like the idea of quickly checking only 'relevant changes', but it simply doesn't work by design. If it's a known design flaw, then Readme should clearly say that. If it's not known - here are steps to reproduce it:

1. Take some Ruby method with max allowed length by your RuboCop config.
2. Add one more line to that method.
3. Run Pronto and RuboCop.

RuboCop will detect violation and mark it at the method definition line. Pronto will not find it as it only checks the diff and the method definition line is not in the diff.

Until this gets resolved, a (slower) correctly working solution is to directly run linters on Jenkins and report new violations via [Violations Plugin](https://wiki.jenkins-ci.org/display/JENKINS/Violations) or some other preferred way.

mmozuras:
Using something like Violations Plugin is a valid alternative, definitely. Depends on your use case!

@sauliusgrigaitis I disagree with `some` but in general, you're right. While Pronto works for most violations, it doesn't work for every violation. But #221 is not the solution 🙂. Here are the options I see:

1. Working with linters creators that they would provide more than just the line number, but the range of lines for a violations. Some linters already do that, but we don't always take advantage. So, in the case you've wrote about: instead of providing the line number for the method definition, all the line numbers for the method would be provided. This would allow Pronto (and other similar tools) to work for these kind of violations too 🙂.

2. Documenting cases that don't work for each runner.

What do you think? What other alternatives do you see?

mmozuras:
For reference, an instance of the same issue: https://github.com/mmozuras/pronto-scss/issues/7

/cc @ibrahima

sauliusgrigaitis:
Well, it depends on the point of view. I think that those code structure related violations (methods complexity, length etc) are way more important that those Pronto are able to detect. That's why I use `some`. Wording for sure is a subjective thing, but I would not use word `perfect` at this stage :)

In any case, I strongly suggest to clearly point this issue in the Readme. A lot of folks may even not notice that a lot of important violations are not reported.

One of the ways to fix it would be to add different levels:
* Current way - works fast, detects only some of violations
* Runs linters on both current and target branch for all changed files. Compares violations and reports new valuations in current branch. A bit slower, finds almost all violations, doesn't detect similar code among separate files.
* Runs linters on both current and target branch for all files. Works slow, detects everything.

This approach doesn't depend on linters efforts to support such tools. I think it's the way to go for the short term.

[...]

PR #221, opened 2017-04-08
Make more clear about the issue #220 in the Readme

sauliusgrigaitis:
Something like this would be more clear that Pronto can't find all the violations.


mmozuras:
@sauliusgrigaitis I think that this kind of message would just create confusion. Ideally, it should explain when it happens. Also, `may` or `may not` depends on specific runner implementation. I'll close this PR and offer more suggestion in #220 for next actions 😄.

PR #228, opened 2017-04-24
Adds a --config option to pass a path to a .pronto.yml config file

### LINKED ISSUE #226: Add an option to pass a .pronto.yml config file
In our current setup we use a Docker image to run pronto in a container to run the checks in our CI. Since we have a lot (120+ projects) that should be checked we want to centralise as much of the configs as possible. Because it isn't possible to set the path of a `.pronto.yml` config file (unless I'm overlooking it?) it means the easiest way to use it is to embed it in all our projects.

It looks like the code already is able to support it, but it just needs to get passed to it? (https://github.com/mmozuras/pronto/blob/master/lib/pronto/config_file.rb#L38)

jeroenj:
closes #226

### DIFF
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -16,6 +16,7 @@ module Pronto
 
     method_option :config,
                   type: :string,
+                  default: ENV['PRONTO_CONFIG'],
                   desc: 'Uses the specified config file.'
 
     method_option :'exit-code',
@@ -50,7 +51,7 @@ module Pronto
                   desc: "Pick output formatters. Available: #{::Pronto::Formatter.names.join(', ')}"
 
     def run(path = nil)
-      ENV['PRONTO_CONFIG'] ||= options[:config]
+      ::Pronto::Config.instance.path = options[:config]
 
       gem_names = options[:runner].any? ? options[:runner] : ::Pronto::GemNames.new.to_a
       gem_names.each do |gem_name|
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -14,6 +14,11 @@ module Pronto
 
     desc 'run', 'Run Pronto'
 
+    method_option :config,
+                  type: :string,
+                  default: '.pronto.yml',
+                  desc: 'Uses the specified config file.'
+
     method_option :'exit-code',
                   type: :boolean,
                   desc: 'Exits with non-zero code if there were any warnings/errors.'
@@ -46,6 +51,8 @@ module Pronto
                   desc: "Pick output formatters. Available: #{::Pronto::Formatter.names.join(', ')}"
 
     def run(path = nil)
+      ENV['PRONTO_CONFIG'] ||= options[:config]
+
       gem_names = options[:runner].any? ? options[:runner] : ::Pronto::GemNames.new.to_a
       gem_names.each do |gem_name|
         require "pronto/#{gem_name}"
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -16,7 +16,6 @@ module Pronto
 
     method_option :config,
                   type: :string,
-                  default: '.pronto.yml',
                   desc: 'Uses the specified config file.'
 
     method_option :'exit-code',
--- a/lib/pronto/config_file.rb
+++ b/lib/pronto/config_file.rb
@@ -1,6 +1,7 @@
 module Pronto
   class ConfigFile
     DEFAULT_MESSAGE_FORMAT = '%{msg}'.freeze
+    DEFAULT_FILE_PATH = '.pronto.yml'.freeze
 
     EMPTY = {
       'all' => {
@@ -35,8 +36,12 @@ module Pronto
       'format' => DEFAULT_MESSAGE_FORMAT
     }.freeze
 
-    def initialize(path = '.pronto.yml')
-      @path = path
+    def initialize
+      @path = if File.exist?(ENV['PRONTO_CONFIG'].to_s)
+                ENV['PRONTO_CONFIG']
+              else
+                DEFAULT_FILE_PATH
+              end
     end
 
     def to_h
--- a/lib/pronto/config_file.rb
+++ b/lib/pronto/config_file.rb
@@ -37,11 +37,9 @@ module Pronto
     }.freeze
 
     def initialize
-      @path = if File.exist?(ENV['PRONTO_CONFIG'].to_s)
-                ENV['PRONTO_CONFIG']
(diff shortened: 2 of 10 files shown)


jeroenj:
@mmozuras I'd be happy to incorporate any feedback. :)

lib/pronto/config_file.rb
```diff
@@ -35,8 +36,12 @@ class ConfigFile
       'format' => DEFAULT_MESSAGE_FORMAT
     }.freeze
 
-    def initialize(path = '.pronto.yml')
-      @path = path
+    def initialize
+      @path = if File.exist?(ENV['PRONTO_CONFIG'].to_s)
```
doomspork: I think we should raise an error and not silently use the default:

```ruby
def initialize
  @path = ENV.fetch('PRONTO_CONFIG', DEFAULT_FILE_PATH) 
  raise "configuration file `#{@path}` missing" unless File.exists?(@path)
end
```

spec/pronto/config_file_spec.rb
```diff
@@ -33,9 +33,41 @@ module Pronto
       context 'only global excludes in file' do
         before do
           File.should_receive(:exist?)
+            .and_return(false)
```
doomspork: Should these `File.should_receive(:exist?)` calls specify their filename?

lib/pronto/config_file.rb
jeroenj: Because of https://github.com/mmozuras/pronto/pull/228/files#diff-d06fbb141ccb6c55a35f4ed5668f6752R19 `PRONTO_CONFIG` will always have a default value. That mens that if no config file exists (which isn't unusual) it would raise an error. But I agree that if it's set to a different value and the file does not exist that it should raise an error.

I didn't think about `Hash#fetch`. I'll update the code. 👍

jeroenj: I've changed it including raising an error when the given path does not exist.

lib/pronto/config_file.rb
```diff
@@ -35,8 +36,10 @@ class ConfigFile
       'format' => DEFAULT_MESSAGE_FORMAT
     }.freeze
 
-    def initialize(path = '.pronto.yml')
-      @path = path
+    def initialize
+      @path = ENV.fetch('PRONTO_CONFIG', DEFAULT_FILE_PATH)
+      return if @path == DEFAULT_FILE_PATH || File.exist?(@path)
```
doomspork: Is this check really necessary?  If the custom file is missing or the default is missing, we should still raise an error.

jeroenj: `ENV['PRONTO_CONFIG']` will always be set because it's the default value of the `--config` option in `Pronto::CLI`.

An option could be to drop the default value for the `--config` option and keep the same logic in this code to fall back to `.pronto.yml` (like you're suggesting).

doomspork: @jeroenj I understand it will be set with a value but we should still ensure that the value (the default `.pronto.yml`) is present.  In the code you have we don't bother checking for `.pronto.yml` we just assume it's there.  In the event that isn't, we won't catch it and instead an error will be raised elsewhere.

jeroenj: Since it will always be set it would mean that `.pronto.yml` would always be required if no custom file path is set. That means that every project should have such a config file.

It should only raise an error if a `--config` or `PRONTO_CONFIG` is used. In case it is not and pronto falls back to `.pronto.yml` (which is the default behavior) it should not raise an error if that file does not exist.

lib/pronto/cli.rb
```diff
@@ -14,6 +14,11 @@ def is_thor_reserved_word?(word, type)
 
     desc 'run', 'Run Pronto'
 
+    method_option :config,
+                  type: :string,
+                  default: '.pronto.yml',
+                  desc: 'Uses the specified cofnig file.'
```
nbekirov: There's a typo in the word `cofnig`.

nbekirov: There's already a `File.exist?` check in the `to_h` method that you might be interested in... 😃

lib/pronto/config_file.rb
nbekirov: There's already a `File.exist?` check in the `to_h` method that you might be interested in... 😃

lib/pronto/cli.rb
jeroenj: Cood catch. My hands tend to get out of sync. :(

Fixed. 👍

lib/pronto/config_file.rb
jeroenj: @nbekirov yeah. I think it should still be in there to check if the default file (`.pronto.yml`) exists...

@doomspork I've removed the default option for `--config` and changed the check to always raise an error if it (or the `PRONTO_CONFIG` env variable) is set. If not it continues and sets it to `.pronto.yml` which is handled in the `File.exist?` check @nbekirov is talking about.

mmozuras:
> Not very happy with passing it as an ENV variable but I saw no easy/clean way to do it otherwise.

Let's try to think of alternatives before committing to passing values via ENV variables. This would require more refactoring, but maybe the whole `Config` class could be global, initialised only once?

jeroenj:
> Let's try to think of alternatives before committing to passing values via ENV variables. This would require more refactoring, but maybe the whole `Config` class could be global, initialised only once?

I think that would make sese. :) 👍

doomspork:
What was the plan with this PR?  Are we going to come up with a new approach or should this be closed for the time being?

jeroenj:
Yes, I'm still planning on continuing to work on it but I had a lot on my plate the last few months which resulted in not enough time for this. I'm on holiday right now but I plan to pick this up again mid-August. Unless someone else picks in of course. :)

jeroenj:
> Let's try to think of alternatives before committing to passing values via ENV variables. This would require more refactoring, but maybe the whole `Config` class could be global, initialised only once?

@mmozuras I've made the `Config` a singleton object to make it global. There are still some pending tests but I've pushed the (working) code. Would you mind having a look at it already?

doomspork:
ping @jeroenj — any progress here?

jeroenj:
I didn't work on it anymore since then. I just (quickly) reread this. I think it would be best if @mmozuras or you gave feedback on the current implementation. Once we settle on an implementation I'll finish the tests so this can be merged. 👍 

(I've been using this code in a fork since then actually.)

ashkulz:
I'm closing this since #435 was merged, please reopen and fix merge conflicts if you still think this should be merged.

PR #231, opened 2017-04-26
Pull Request Review "Accept" header needed

ifokeev:
https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,10 +105,16 @@ Then just run it:
 $ PRONTO_GITHUB_ACCESS_TOKEN=token pronto run -f github -c origin/master
 ```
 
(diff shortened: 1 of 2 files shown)


mmozuras:
@ifokeev hm, I tried it again and it works for me without a header. Wonder why that is 🤔

Thanks for the contribution!

mmozuras:
Released as part of [v0.9.2](https://github.com/mmozuras/pronto/releases/tag/v0.9.2).

npetkov:
@mmozuras The custom media type shouldn't be needed any more, just 10 days after the merge it [became official](https://developer.github.com/changes/2017-05-09-end-black-cat-preview/).

Issue #227, opened 2017-04-23
Converting to a GitHub organisation


mmozuras:
Hey,

I've decided to consider moving all Pronto related repositories to an organisation. Before I decide to pull the trigger, I'd like to ask for some input. I personally think that it would help organise, maintain and contribute in a better way ✊.

First thing that I've checked:
* [x] Github will automatically redirect the old personal repo to the new org repo.

@mknapik, @doomspork, @jeroenj, @aergonaut, @nicka, @andrey-skat, @nbekirov, @gabealmer: all of you have contributed significant chunks of code to Pronto, so your input would be especially valuable 🙇.

doomspork:
Howdy @mmozuras!  I see no problems with switching to an organization so long as you're happy to do so 😀 

Let me know if there's anything I can do to help.

aergonaut:
Yep I see no problems doing this!

nicka:
Same here, please go ahead 👍

jeroenj:
Sounds great to me. As long as you're okay with it. :)

nbekirov:
That sounds like a great idea. 👍

mmozuras:
Thanks for support! 🙂

And here are some initial ideas for the logo: [1](https://cloud.githubusercontent.com/assets/54526/25452313/e0db7178-2acd-11e7-8757-2db063fdafc3.jpg), [2](https://cloud.githubusercontent.com/assets/54526/25452316/e1035102-2acd-11e7-9cf4-f1a6c4fa7f16.jpg), [3](https://cloud.githubusercontent.com/assets/54526/25452315/e1033cbc-2acd-11e7-8717-75aebd26b8e6.jpg), [4](https://cloud.githubusercontent.com/assets/54526/25452314/e1022840-2acd-11e7-8875-c00968f6b85b.jpg). Drawn up by a good friend. Which do you like? Which conveys Pronto's idea in the best way? Any other ideas on what could be Pronto's logo?

doomspork:
@mmozuras I like #2.   I'll ponder logo ideas over the next few days.  Thanks again for including all of us in the decision making process 😁

aergonaut:
Yeah, I like 2 the most. Conveys the theme of speed the best in my opinion.

giedriusr:
2nd

jeroenj:
I like the first one the most. But I really like the second one too. :)

nbekirov:
I really like #2

mmozuras:
[...]

PR #235, opened 2017-05-08
[Fix #233] Skip reviewing PR without comments

### LINKED ISSUE #233: GitHub PR Review formatter console warning
Pronto tries to submit a request to GitHub even if there are no comments to submit.

This results in error message posted to console:

pronto_zero_case

ivanovaleksey:
Hi @mmozuras,

I tried to fix #233.

Not sure about spec. It seems like `spec/pronto/github_spec.rb` uses RSpec `context`s to describe behaviour. And I believe `it` is supposed to do it.

Please take a look.

### DIFF
--- a/lib/pronto/github.rb
+++ b/lib/pronto/github.rb
@@ -39,6 +39,8 @@ module Pronto
     end
 
     def create_pull_request_review(comments)
+      return if comments.empty?
+
       options = {
         event: 'COMMENT',
         accept: 'application/vnd.github.black-cat-preview+json', # https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review
--- a/spec/pronto/github_spec.rb
+++ b/spec/pronto/github_spec.rb
@@ -124,5 +124,57 @@ module Pronto
         end
       end
     end
+
+    describe '#create_pull_request_review' do
+      subject { github.create_pull_request_review(comments) }
+
+      let(:octokit_client) { double(Octokit::Client) }
+
+      before do
+        github.instance_variable_set(:@client, octokit_client)
+      end
+
+      context 'with no comments' do
+        let(:comments) { [] }
+
+        specify do
+          octokit_client.should_not_receive(:create_pull_request_review)
+          subject
+        end
+      end
+
+      context 'with comments' do
+        before do
+          github.should_receive(:pull_id).once.and_return(pull_id)
+        end
+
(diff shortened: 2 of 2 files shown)


spec/pronto/github_spec.rb
```diff
@@ -124,5 +124,57 @@ module Pronto
...
+
+        specify do
+          octokit_client.should_not_receive(:create_pull_request_review)
+          subject
+        end
+      end
+
+      context 'with comments' do
+        before do
+          github.should_receive(:pull_id).once.and_return(pull_id)
```
ivanovaleksey: Is it OK to stub this method and not to deal with `ENV['PRONTO_PULL_REQUEST_ID']`?

lib/pronto/github.rb
```diff
@@ -39,6 +39,8 @@ def create_pull_comment(comment)
     end
 
     def create_pull_request_review(comments)
+      return if comments.empty?
```
ivanovaleksey: Another place for adding this could be `GithubPullRequestReviewFormatter#submit_comments`.
Not sure which one should be used.

spec/pronto/github_spec.rb
mmozuras: @ivanovaleksey yup, I think that's fine.

lib/pronto/github.rb
mmozuras: @ivanovaleksey I think that it makes sense here 👍

ivanovaleksey: Yes, that was what I thought about.

SuperTux88:
This is now fixed since 1,5 month, can we please have a release for this?

denschub:
@mmozuras ping?

mmozuras:
@SuperTux88 @denschub sorry about that. Had a rough month in my personal life, which made contributing to OSS more than difficult. This again shows why I truly completing #227 should be a priority.

mmozuras:
Released as part of [v0.9.4](https://github.com/prontolabs/pronto/releases/tag/v0.9.4).

denschub:
In that case: sorry for being annoying. Hope you are better and **thank you very much** for your work. It is very appreciated. 😻

mmozuras:
@denschub no worries, I understand your frustration 😄. And thank you for the kind words 🙇.

PR #241, opened 2017-05-30
Add pronto-flow and pronto-dirty_words to runners

kevinjalbert:
Add pronto-flow -- uses flowtype for type checking javascript
Add pronto-dirty_words -- checks for uses of the seven dirty words


mmozuras:
@kevinjalbert wow, thanks! 🙇

PR #242, opened 2017-05-30
Fix #139 - Add article on how to build new runner

### LINKED ISSUE #139: How to write a pronto runner ?
I cannot seem to find docs how to write a new pronto runner and how to run/test it locally. Would be great if i can get some help.

Basically i want to see if i can write a pronto which can look at change and comment if its allowed or not.

kevinjalbert:
Add a new article on how to build a new pronto runner.


mmozuras:
@kevinjalbert 🙇

PR #247, opened 2017-06-16
When HEAD is detached find GitHub PR by SHA 

drewish:
I've been trying to use the GithubPullRequestReviewFormatter with a Jenkins pipeline but ran into a problem. Jenkins doesn't expose the PR number and it checks out the commit from the webhook rather than the branch. The best fix I could come up with was modifying pronto to find the PR based off the SHA if the repo has a detached HEAD.

I wasn't sure if this would be an acceptable change so I didn't bother with any specs. If it's something you'd consider merging I'm happy to add tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,3 +17,6 @@ Metrics/BlockLength:
   ExcludedMethods: Struct.new
   Exclude:
     - spec/**/*.rb
+
+Rubocop/Metrics/AbcSize:
+  Max: 25
--- a/lib/pronto/git/repository.rb
+++ b/lib/pronto/git/repository.rb
@@ -66,6 +66,10 @@ module Pronto
         head.oid
       end
 
+      def head_detached?
+        @repo.head_detached?
+      end
+
       private
 
       def empty_patches(sha)
--- a/lib/pronto/github.rb
+++ b/lib/pronto/github.rb
@@ -92,6 +92,10 @@ module Pronto
                   pull_requests.find { |pr| pr[:number].to_i == env_pull_id }
                 elsif @repo.branch
                   pull_requests.find { |pr| pr[:head][:ref] == @repo.branch }
+                elsif @repo.head_detached?
+                  pull_requests.find do |pr|
+                    pr[:head][:sha] == @repo.head_commit_sha
+                  end
                 end
     end
 
--- a/spec/pronto/github_spec.rb
+++ b/spec/pronto/github_spec.rb
@@ -3,7 +3,7 @@ module Pronto
     let(:github) { described_class.new(repo) }
 
     let(:repo) do
-      double(remote_urls: ['git@github.com:prontolabs/pronto.git'], branch: nil)
+      double(remote_urls: ['git@github.com:prontolabs/pronto.git'], branch: nil, head_detached?: false)
     end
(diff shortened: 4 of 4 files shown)


mmozuras:
@drewish seems good to me 👍

@mknapik @doomspork @aergonaut what do you think?

aergonaut:
Seems alright to me. I don't think it's possible to have more than one PR with the same head SHA, so finding the PR based on that seems fine.

drewish:
Cool I'll throw some specs together. Any suggestions on how to resolve that code climate issue?

drewish:
Actually looking at the existing tests I'm not sure specs for this make a ton of sense. `Pronto::Github#pull` is private and `Pronto::Git::Repository#head_detached?` just delegates to the  `Rugged::Repository`. If there's something that makes sense to the maintainers I'm happy to add it though.

doomspork:
@mmozuras sounds like a good idea, I'd run into this issue myself w/ a previous client.

aergonaut:
Codeclimate is complaining about the ABC size of that block. I personally think the block is fine as written so I would rather just increase the max value of that metric. @mmozuras what do you think about that?

mknapik:
@drewish You might find it useful: https://github.com/u2i/jenkins-pipeline-utils/blob/master/src/me/knapik/GitHub.groovy
https://github.com/u2i/jenkins-pipeline-utils/blob/master/vars/foreachPullRequest.groovy

And an example usage:
```groovy

        foreachPullRequest(GITHUB_ACCESS_TOKEN_CREDENTIALS_ID) { int pullRequestId, String targetBranchRef ->
            withEnv(["PULL_REQUEST_ID=${pullRequestId}"]) {
                rake "ci:pronto:pr[origin/${targetBranchRef}]"
            }
        }
```

drewish:
@mknapik oh neat yeah i've got some ruby code that does something similar. trying to do it in groovy with the CPS stuff was too much of a frustration for me. i found it easier to just call it via the shell. but when i was looking at the code in pronto i realized it just needed a couple tweaks to let me ditch my custom code.

mknapik:
@drewish The name of the job for PR job type is `PR-XX`.
image
So it is possible to strip `PR-`. Not as straightforward though.

drewish:
@mknapik ah thanks, I checked and we're doing the branch based PRs so that makes more sense to me now. 

@mmozuras I can bump the ABC metric as part of this PR if you think that makes sense or I'm happy to do that in a separate PR if that needs to be a prerequisite for this.

mmozuras:
@drewish we're in no hurry, so would prefer if you'd do it as part of this PR 👍

drewish:
Pushed up a change to the rubicon config. It looks like there were a few other places getting flagged by CC for that complexity issue that are now happier.

drewish:
I bumped it up to 25 since that was showing ~21. It looks like we've resolved a few other places that were getting flagged for that complexity issue.

It's looking green!

drewish:
@mmozuras let me know if there's anything else this PR needs.

mmozuras:
@drewish nope, looks good! 👍

mmozuras:
Released as part of [v0.9.4](https://github.com/prontolabs/pronto/releases/tag/v0.9.4).

Issue #240, opened 2017-05-29
"pronto run" doesn't to do anything, --staged crashes


stiller-leser:
I created a bogus file on a bogus branch on my MacOS machine:

```
#I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY LONG LINE I AM A VERY

def i_am_a_vunction lala, lulu
        puts lulu
end

i_am_a_vunction "hahaha"
```
However neither of the following commands yields a result until I have commited the changes:

pronto run (before or after git add .)
pronto run --unstaged (before git add .)

Running pronto run --staged (after git add .) I get the following error message:

```
/Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/git/repository.rb:52:in `new': the path 'test.rb' does not exist in the given tree (Rugged::TreeError)
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/git/repository.rb:52:in `blame'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/git/patch.rb:17:in `blame'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/git/line.rb:45:in `blame'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/git/line.rb:20:in `commit_sha'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-0.9.3/lib/pronto/message.rb:18:in `initialize'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-rubocop-0.9.0/lib/pronto/rubocop.rb:49:in `new'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-rubocop-0.9.0/lib/pronto/rubocop.rb:49:in `new_message'
	from /Users/MyUser/.rvm/gems/ruby-2.3.0/gems/pronto-rubocop-0.9.0/lib/pronto/rubocop.rb:41:in `block (2 levels) in inspect'
[...]

PR #258, opened 2017-08-10
Blame returning nil when file does not exist in the git tree

pettomartino:
Related to https://github.com/prontolabs/pronto/issues/240

Instantiating `Rugged::Blame` with a new file(`Rugged::Repository#status == index_new`) will throw `Rugged::TreeError` with `does not exist in the given tree`.


doomspork:
@pettomartino what is your twitter handle?  😀

pettomartino:
@doomspork same, @pettomartino

durnin:
Hello, is this getting merged anytime soon? I'm also trying to use pronto with pre-commit hook using --staged. Thanks!

doomspork:
@mmozuras any thoughts on this one, is it good to go?

bjornlinder:
I tested out the changes in my pre-commit hook, didn't see the output I was expecting from pronto. Ideally we'd have pronto linting the newly added (staged) files. Here's what I'm going with for now:

```
if (( $(echo "git diff --name-only --diff-filter=A HEAD" | wc -l) > 0 ))
then
  pronto_output=$(pronto run)
else
  pronto_output=$(pronto run --staged)
fi
```

mmozuras:
@doomspork yup, looks good, sorry for taking so long 🙇 

@bjornlinder I think that [this issue](https://github.com/prontolabs/pronto/issues/263) is related to what you wrote about. Would love to read more of your thoughs on how to best approach this 😄.

Issue #254, opened 2017-07-26
Pronto runner for inspec


stiller-leser:
I have written a pronto runner for chef's inspec and test kitchen. You can find it [here](https://github.com/stiller-leser/pronto-inspec)

I'd call it an alpha version, but I'd love feedback on it. What it allows you to do is to specify which inspec suite should be run depending on the files that have been changed. It currently uses `kitchen test` for each suite, but that could be improved in future versions.

So who is this for? If you have a bigger chef repository (read not a git repo for every cookbook) you could use this to run tests just on the cookbook that has been changed. Also, depending on how your repository is set up, it allows you a very granular control over which suites to run when.

Thanks for the good work!

mmozuras:
@stiller-leser thanks for writing it! 🙇 

Could you make a PR, adding it to the list [here](https://github.com/prontolabs/pronto/#runners)? 😄

PR #262, opened 2017-09-11
Adds inspec runner for pronto to runner overview, re. #254

stiller-leser:
This adds my pronto runner for inspec tests to the runner overview in README.md, "solves" #254


doomspork:
Thank you @stiller-leser 😁

Issue #261, opened 2017-09-06
Error handling when PR not found


panjan:
Hi,

When you run `pronto run -f github_pr` and [`Pronto::GitHub#pull`](https://github.com/prontolabs/pronto/blob/master/lib/pronto/github.rb#L90) returns `nil` (e.g. because you're on the wrong branch), the code execution continues. As a result, the following request to GitHub API fails with `404 Not Found` because the request goes to `/repos/{company}/{repo}/pulls/comments` instead of `/repos/{company}/{repo}/pulls/{id}/comments`.

This situation occured in our CI build and it took me quite some time until I figured what was the source of the problem. IMHO it would be useful if, after not finding the PR, the program failed gracefully and gave an informative error message.

I'll be happy to submit a PR. Please let me know if you'd accept it.

mknapik:
So the use case is: `PULL_REQUEST_ID` env var is not set, branch exists both locally and remotely but no PR is submitted for that branch?
Returning informative message sounds reasonable to me.

panjan:
OK, I have a PR for this. Would be grateful for a review. Thanks! 🙏

ashkulz:
Mass-closure of old issues more than 4 years old, please reopen and/or comment if this is still valid.

PR #269, opened 2017-09-30
Raise error when Github PR not found

panjan:
related: #261

### DIFF
--- a/spec/pronto/github_spec.rb
+++ b/spec/pronto/github_spec.rb
@@ -46,11 +46,6 @@ module Pronto
       context 'three requests for same comments' do
         specify do
           Octokit::Client.any_instance
-            .should_receive(:pull_requests)
-            .once
-            .and_return([])
-
-          Octokit::Client.any_instance
             .should_receive(:pull_comments)
             .with('prontolabs/pronto', 10)
             .once
@@ -66,6 +61,8 @@ module Pronto
 
       context 'pull request does not exist' do
         specify do
+          ENV['PRONTO_PULL_REQUEST_ID'] = 'foo'
+
           Octokit::Client.any_instance
             .should_receive(:pull_comments)
             .and_raise(Octokit::NotFound)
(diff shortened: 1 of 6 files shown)


doomspork:
@panjan this looks like a good start but the build is not passing, could you pick that up first?

spec/pronto/github_spec.rb
```diff
@@ -66,6 +61,8 @@ module Pronto
 
       context 'pull request does not exist' do
         specify do
+          ENV['PRONTO_PULL_REQUEST_ID'] = 'foo'
```
mknapik: I think the tests should not mutate global state to avoid random failures dependent on test order.
I'd either cache the old result:

```ruby
let(:old_env_var) { ENV['PRONTO_PULL_REQUEST_ID'] }
let(:env_var) { 'foo' }
before(:each) do
  old_env_var
  ENV['PRONTO_PULL_REQUEST_ID'] = env_var
end
after(:each) do
  ENV['PRONTO_PULL_REQUEST_ID'] = old_env_var
end
```

or use stubs:

```ruby
allow(ENV).to receive(:[]).with('PRONTO_PULL_REQUEST_ID').and_return('foo')
allow(ENV).to receive(:fetch) ....
```

panjan: Used stubs like so: `ENV.stub(:[]).with('PRONTO_PULL_REQUEST_ID').and_return(10)`

Couldn't use `allow` as the "should syntax" is configured. 

Also had to stub other calls to `ENV` using `ENV.stub(:[])`. That's because ENV is called like 25 times in [Pronto::Config](https://github.com/prontolabs/pronto/blob/master/lib/pronto/config.rb#L10). This leads me to an idea: Wouldn't it be better to create method `Pronto::Config#pronto_pull_request_id` and use and stub that?

doomspork:
@prontolabs/core any final thoughts on this?

Schwad:
@mmozuras @mknapik looks like things are passing over here, any other needed tweaks or is this ready for rollout? :)

PR #290, opened 2018-04-01
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

note: some people don't like changing changelogs, others view that while a commit may be flawed, there's no reason for most people to suffer from the errors. I can trivially drop the commit if requested. (I can also squash if requested.)


doomspork:
Thank you @jsoref 👍

PR #292, opened 2018-04-17
Convert `PRONTO_MAX_WARNINGS` to Integer

fwitzke:
**What does this do?**
It converts `PRONTO_MAX_WARNINGS` from String to Integer.

**Why?**
ENV variables are always read as Strings. So when trying to set this from an environment variable an exception is raised in [`Pronto::Runners#exceeds_max?`](https://github.com/prontolabs/pronto/blob/master/lib/pronto/runners.rb#L37) since it tries to compare an Integer to a String.

**How to test this?**
The issue can be reproduced on the latest version simply by setting max warnings through ENV:
```shell
⇒  PRONTO_MAX_WARNINGS=1 pronto run
ArgumentError: comparison of Integer with String failed
```

### DIFF
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -40,7 +40,7 @@ module Pronto
     end
 
     def max_warnings
-      ENV['PRONTO_MAX_WARNINGS'] || @config_hash['max_warnings']
+      ENV['PRONTO_MAX_WARNINGS'] && ENV['PRONTO_MAX_WARNINGS'].to_i || @config_hash['max_warnings']
     end
 
     def message_format(formatter)
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -40,7 +40,7 @@ module Pronto
     end
 
     def max_warnings
-      ENV['PRONTO_MAX_WARNINGS'] && ENV['PRONTO_MAX_WARNINGS'].to_i || @config_hash['max_warnings']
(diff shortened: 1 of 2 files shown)


lib/pronto/config.rb
```diff
@@ -40,7 +40,7 @@ def bitbucket_hostname
     end
 
     def max_warnings
-      ENV['PRONTO_MAX_WARNINGS'] || @config_hash['max_warnings']
+      ENV['PRONTO_MAX_WARNINGS'] && ENV['PRONTO_MAX_WARNINGS'].to_i || @config_hash['max_warnings']
```
mikegee: Consider
```ruby
ENV['PRONTO_MAX_WARNINGS'] && Integer(ENV['PRONTO_MAX_WARNINGS']) || @config_hash['max_warnings']
```
so only integer values in `ENV['PRONTO_MAX_WARNINGS']` will be allowed.

fwitzke: I like it, updating the PR.

mknapik:
Are `pronto` tests using real API?
@prontolabs/core Why are we getting `Octokit::TooManyRequests` randomly?

doomspork:
@mknapik it looks like they're mocked but something is clearly afoot: https://github.com/prontolabs/pronto/blob/master/spec/pronto/github_spec.rb#L48

PR #301, opened 2018-06-27
Add bitbucket auto approve pull request feature

starchow:
I'm using bitbucket and pronto to auto check style violations in pull request, but there are no way to know if a pull request is checked or not. So I added an approve option to let us know which pull request is checked.

Just set auto_approve: true 

Screenshot

And here is the result 

Screenshot

### DIFF
--- a/lib/pronto/bitbucket.rb
+++ b/lib/pronto/bitbucket.rb
@@ -32,6 +32,13 @@ module Pronto
       end
     end
 
+    def approve_pull_request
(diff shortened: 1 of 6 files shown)


starchow:
@mmozuras yes, it is a signal to let us know that pronto has been run. 
The check will run again and take off approval if there are new messages.

mmozuras:
@starchow 👍

PR #305, opened 2018-07-17
Detect file renames correctly. Fixes #218

### LINKED ISSUE #218: pronto having issues with renamed files
I ran into issues with pronto where it was commenting on commits which were outside of the differences using the --commit flag

Steps - 
1. Basically have a file in your repo /Test/SomeFile.php
2. Add in some generic php code which contains some lint errors
3. Check into gitlab
4. Add in several more additions to that file which pass lint (pronto will see no errors)
5. Rename that file to /Test/SomeFile2.php
- at this point pronto will lint against everything in the file, including all commits back which really should not be in the scope of the diff

bit of background on what we are doing in our gitlab-ci:
pronto run --runner phpcs --formatters gitlab --exit-code --commit=$(git --no-pager log --pretty=format:%H HEAD^1 --reverse | tail -1)

basically we do a diff of the previous commit to compare against. It seems a rename operation will treat the entire file as different and will comment on every linting error (in our case it commented on commits which were several months back prior to us adding in pronto).

siebertm:
Before this change, when renaming files, pronto treated it as a whole
new file, complaining about old code smells. By calling `#find_similar!`
on the diff, rugged/libgit2 checks for these renames and removes them
from the diff. This leaves only new lines in the diff and we can
correctly report on new problems only.

## Help needed: Testing.

In order to be able to test this code, I was planning to create a commit on top of the `test.git` repository which does the rename. Doing this would mean that all the other tests would also have to be adapted (as the `head` changed). I wanted to create a branch for that, but could not set it up correctly so that the test picks the correct diff (it was always empty). 

So I resorted to copy the whole repository which is a bit much for this one test. Maybe somebody can support me here on how to test that correctly? I wasn't able to find any documentation about this part of my contribution...


siebertm:
(tests fail due to rate limiting afaics)

mmozuras:
@siebertm thanks and sorry for reviewing it so late. I've merged [this](https://github.com/prontolabs/pronto/pull/329) instead, but only because it was newer! 🙇

PR #309, opened 2018-08-07
Update rainbow and solve conflict with rubocop dependencies

Schweigert:
Hi guys,

This PR resolves compatibility issues with rubocop. The rubocop needs the rainbow in the version> = 2.2.2. It is conflicting between the last version of the pronto and the rubocop when installing the gem in a project. In addition, the latest version published in rubygems requires version 2.1 exactly, and it's broken with rubocop (>= 2.2.2).

firefox_screenshot_2018-08-07t21-26-04 813z Runtime dependence of pronto.

[Rubocop dependence](https://github.com/rubocop-hq/rubocop/blob/master/rubocop.gemspec#L41)

This PR uses the same version of the lasted version of the rainbow gem from the rubocop gem.
:smile_cat:

### DIFF
--- a/pronto.gemspec
+++ b/pronto.gemspec
@@ -43,7 +43,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency('gitlab', '~> 4.0', '>= 4.0.0')
   s.add_runtime_dependency('httparty', '>= 0.13.7')
   s.add_runtime_dependency('octokit', '~> 4.7', '>= 4.7.0')
-  s.add_runtime_dependency('rainbow', '>= 2.2', '< 4.0')
+  s.add_runtime_dependency('rainbow', '>= 2.2.2', '< 4.0')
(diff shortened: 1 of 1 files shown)


doomspork:
Hi @Schweigert could you explain this change and the problem a bit?  It's not immediately clear to me how adding patch level to the deps would fix anything.

Schweigert:
@doomspork The rubocop needs the rainbow in the version> = 2.2.2. It is conflicting between the last version of the pronto and the rubocop when installing the gem in a project. In addition, the latest version published in rubygems requires version 2.1 exactly, and it's broken with rubocop (>= 2.2.2).

firefox_screenshot_2018-08-07t21-26-04 813z Runtime dependence of pronto.

[Rubocop dependence](https://github.com/rubocop-hq/rubocop/blob/master/rubocop.gemspec#L41)

PR #287, opened 2018-03-06
Bump rainbow 3.0

dylanpinn:
Fixes #282

PR #311, opened 2018-08-17
Add pronto-goodcheck

aergonaut:
Add link to [pronto-goodcheck](https://github.com/aergonaut/pronto-goodcheck).

Note: This depends on the [unreleased change to allow rainbow 3.0](https://github.com/prontolabs/pronto/pull/287), so this shouldn't be merged until that change is released to avoid confusing people. I think the version conflict can be ignored if using Bundler and pronto is specified as a Git dependency instead of from RubyGems. But I think we would rather have people using RubyGems, and this also doesn't work for someone who isn't using Bundler.

cc @prontolabs/core


doomspork:
Thank you @aergonaut!

aergonaut:
Thanks @doomspork. Yup, I hope we can get a new version cut soon!

PR #314, opened 2018-09-21
Add support OAuth2 and consumers authentication

one0fnine:
Bitbucket Cloud v1 APIs are deprecated

Bitbucket Cloud REST API version 1 is deprecated effective 30 June 2018. All 1.0 APIs will be removed from the REST API permanently on 31 December 2018.


ashkulz:
I think this is not relevant as the BitBucket API 2.0 is currently used?

PR #321, opened 2018-11-13
Add pronto-blacklist runner to readme

pbstriker38:
pronto-blacklist is useful for blocking additions of blacklisted strings. Such as classes that are being deprecated.


mknapik:
Thanks for your contribution!

Issue #328, opened 2018-12-28
Stop Considering renamed-only Files as Additions


alessio-signorini:
The current implementation of `Git::Repository` does not handle correctly file renames. These are the considered **additions** and create numerous issues when trying to submit Github PRs. A few of the linked issues I could find are:
1. Pronto doesn't handle file renames well (#69)
2. 422 - Validation Failed (#149)
3. pronto having issues with renamed files (#218)
4. Issue with renamed files and github_pull_request_review_formatter (#259)
5. Pronto is failing to post reviews to github pull request (#284)

PR #329, opened 2018-12-28
Correctly Handle renamed-only Files

alessio-signorini:
The current implementation of Pronto does not correctly consider renamed-only files which are considered new additions. This forces the runners to analyze the code and creates numerous issues when submitting PRs to Github. A summary of the issue and related issues can be found in #328.

This PR addresses the issue using `patches.find_similar!(:renames => true)` so that `patches` have the correct type (i.e., `renamed`) for those files. The analysis will be slightly slower but it should not be a problem.


alessio-signorini:
I included a test! It adds a bunch of fixtures files (due to the extra git repository). Sorry about that.

The only change in the actual code is [this line](https://github.com/prontolabs/pronto/pull/329/files#diff-cb2518bc21b8fd7529e84e90fa9807bcR19).

lib/pronto/git/repository.rb
```diff
@@ -16,6 +16,7 @@ def diff(commit, options = nil)
                           else
                             merge_base = merge_base(commit)
                             patches = @repo.diff(merge_base, head, options)
+                            patches.find_similar!(:renames => true)
```
doomspork: I think this might read better expressed in the JSON format, thoughts?

```suggestion
                            patches.find_similar!(renames: true)
```

spec/fixtures/renamed-file.git/hooks/applypatch-msg.sample
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
```
doomspork: Are all of these hook files in the fixtures entirely necessary for the test to fun? If at all possible to prone them down, we should 👍

alessio-signorini: I was just doing that. It bothered me so much to see all those extra unnecessary files! :)

lib/pronto/git/repository.rb
alessio-signorini: Definitely, that conforms to the new Ruby style guide. Changed it.

alessio-signorini:
Should be all good now @doomspork, thanks for the review!

mmozuras:
@alessio-signorini concur with @doomspork - awesome job 😄

Issue #330, opened 2018-12-28
Do Not Require to run From a Git Repository


alessio-signorini:
The current implementation forces the user to run `pronto run` from a Git repository. This can be inconvenient and creates issues when the gems can not be added to the local Gemfile due to conflicts with others in the project. For example, `foreman-0.85.0` requires `thor-0.19.1` and will not allow `pronto-0.9.5` to install.

PR #331, opened 2018-12-28
Allow to Specify PATH in "run" command

alessio-signorini:
This patch addresses #330 and allows to use a **path** in `pronto run` as in
```bash
pronto run ~/myproject
```
it also works with a specific file inside that project, e.g.,
```bash
pronto run ~/myproject/test/first_test.rb
```
Unfortunately I did not add any test because I could not find any for the CLI.

### DIFF
--- a/lib/pronto/cli.rb
+++ b/lib/pronto/cli.rb
@@ -70,7 +70,7 @@ module Pronto
         exit(error_messages_count)
       end
     rescue Rugged::RepositoryError
-      puts '"pronto" should be run on/from a git repository'
+      puts '"pronto" must be run from within a git repository or must be supplied the path to a git repository'
     rescue Pronto::Error => e
       $stderr.puts "Pronto errored: #{e.message}"
     end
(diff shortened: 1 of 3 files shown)


lib/pronto/cli.rb
```diff
@@ -56,16 +58,19 @@ def run(path = nil)
...
+        file = relative.length != path.length ? relative : nil
+        ::Pronto.run(commit, '.', formatters, file)
       end
       if options[:'exit-code']
         error_messages_count = messages.count { |m| m.level != :info }
         exit(error_messages_count)
       end
     rescue Rugged::RepositoryError
-      puts '"pronto" should be run from a git repository'
+      puts '"pronto" should be run on/from a git repository'
```
mikegee: "on/from" seems weird to me, but I know what you mean.

Consider these more succinct or verbose options:

>  "pronto" should target a git repository

> "pronto" must be run from within a git repository or must be supplied the path to a git repository

mmozuras:
@alessio-signorini this is amazing, great effort! 🙇

Issue #324, opened 2018-12-28
GithubSpec Test Fails (Pronto::Github#create_pull_request_review with comments)


alessio-signorini:
Master branch tests are failing with the following error
```ruby
1) Pronto::Github#create_pull_request_review with comments 
     Failure/Error:
       octokit_client
         .should_not_receive(:create_pull_request_review)
         .with('prontolabs/pronto', pull_id, options)
         .once
     
     RuntimeError:
       `count` is not supported with negative message expectations
     # ./spec/pronto/github_spec.rb:170:in `block (4 levels) in <module:Pronto>'
```

PR #332, opened 2018-12-29
Remove Ruby 2.2

doomspork:
Simple change to get the build passing. cc @alessio-signorini


doomspork:
Build fails for other reasons, #324 will address those failures. This just removes an unsupported Ruby as pointed out by @alessio-signorini

PR #320, opened 2018-11-01
Change Github review comment mode to "REQUEST_CHANGES"

rnitta:
# Issue
https://github.com/prontolabs/pronto/issues/319

# Implements
Just changed `event` from 'COMMENT' to 'REQUEST_CHANGES',  which is submitted with github review comment.

Issue #319, opened 2018-10-31
Github review comment should be "REQUEST_CHANGES" mode


rnitta:
https://github.com/prontolabs/pronto/blob/899c6f270ed80cce1585abf1cf1b0006899021f3/lib/pronto/github.rb#L52

Now it is submitted with "COMMENT" mode.
it should be "REQUEST_CHANGES" (and this is also semantic).

Or, the mode should be selectable by user.

thoughts?

see also: 
https://octokit.github.io/octokit.rb/Octokit/Client/Reviews.html#create_pull_request_review-instance_method
https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review

mikegee:
seems legit to me 👍

mknapik:
I like the idea of submitting as Request Changes but making it default and only option seems to be pretty arbitrary.
I think it would be better to make the mode configurable so we don't impose any particular workflow on users.

mmozuras:
I agree with @mknapik - "request changes" makes sense as the default and I think that we should be it configurable.

ashkulz:
This has been implemented.

PR #334, opened 2019-03-19
Add configurable Github review_type with 'REQUEST_CHANGES' as default

gyfis:
NOTE: This PR is a replacement for PR #320 which lost its traction.

## Issue
#319 

## Implements
Similarly to #320 this adds `REQUEST_CHANGES` as the default event, with some more code around the configurability and specs for that.

### DIFF
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -23,10 +23,10 @@ module Pronto
         ENV['PRONTO_GITHUB_REVIEW_TYPE'] ||
         @config_hash.fetch('github_review_type', false)
 
-      if review_type == 'comment'
-        'COMMENT'
-      else
+      if review_type == 'request_changes'
         'REQUEST_CHANGES'
+      else
+        'COMMENT'
       end
     end
 
--- a/lib/pronto/config.rb
+++ b/lib/pronto/config.rb
@@ -18,6 +18,18 @@ module Pronto
       consolidated
     end
 
+    def github_review_type
+      review_type =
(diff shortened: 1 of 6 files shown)


doomspork:
@gyfis sorry for the delay on getting you feedback. Could you rebase against master so we can get this merged?

gyfis:
@doomspork Thanks for the reminder, I rebased the code against the latest master.

mknapik:
Looks solid. I have doubts about `REQUEST_CHANGES` as a default behavior. As much as I'd prefer to keep it as `REQUEST_CHANGES`, I'm afraid some of our users might be annoyed that we're breaking compatibility.

@gyfis @prontolabs/core What do you think?

doomspork:
Agreed @mknapik. @gyfis can we swap the default behavior for now?

gyfis:
@mknapik That makes sense. Thanks for the review!

@doomspork Updated the code to use COMMENT as the default.

doomspork:
Thank you @gyfis! @mknapik any final thoughts before we merge?

gyfis:
@doomspork @mknapik How do you feel about this change?

doomspork:
Looks good @gyfis, thank you for your patience! We're definitely in need of more maintainers if you'd like to be more involved in Pronto's development 😁

gyfis:
Thanks for the merge @doomspork and for the offer! I'm busy (as we all are), but I'll keep it in mind 👍

PR #337, opened 2019-04-16
readme: add a pointer to a new runner built to run 'hlint' linter

paulRbr:
We had the need to be able to run `hlint` in our codebase as a haskell
programmers so we created the [`pronto-hlint`](https://github.com/fretlink/pronto-hlint/)
 gem.

This PR adds a mention in the readme of Pronto.


doomspork:
Thank you @paulRbr! Always love seeing new runners ❤️

PR #343, opened 2019-05-13
Make readme describe github access token usage correctly

matt3play:
It's possible that I'm missing something here, but I think the correct environemnt variable for the github access token is GITHUB_ACCESS_TOKEN, not PRONTO_GITHUB_ACCESS_TOKEN.  I've updated the readme accordingly.


matt3play:
Oops.  Using an old version of pronto.   Closing PR.

PR #354, opened 2019-08-05
Bitbucket deprecated its API v1.0: update all endpoints to API v2.0

bradrees:
Merging of 3 pull requests #347, #348 and #352, plus my changes to update the endpoints.

The only significant change here is to use the `commit` endpoint instead of the removed `changeset` endpoint. This appears to be working for me, and I'm getting the comments on my commit as per before the v2.0 change.

@doomspork, @starchow please take a look, thanks!

https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Busername%7D/%7Brepo_slug%7D/commit/%7Bnode%7D/comments#post

PR #363, opened 2019-11-07
fix: usr: 362: fix duplicate comments issue in bitbucket

tabraiz12:
Bitbucket v2 api fetches only 10 records and it is paginated, so its  not able to check beyond 10 comments and duplicates the comments


doomspork:
@tabraiz12 are these changes still necessary with the merge of #354 ?

bradrees:
@doomspork these changes should not be required now, that PR you accepted does handle the paging

doomspork:
Thank you @bradrees

Issue #365, opened 2019-11-14
Add option to exclude a runner


julianrubisch:
This requires some explanation. IIUC you can specify which runners to run with `-r`, but pronto will actually use this to `require` a certain runner.

I wrote a `ruby-standardrb` runner https://github.com/julianrubisch/pronto-standardrb (see respective PR #364) basically only by subclassing `pronto-rubocop` (because that's what `standard` is, an opinionated wrapper around rubocop).

Now of course in that implementation I need to `require` the rubocop runner: https://github.com/julianrubisch/pronto-standardrb/blob/c5669db79609f4fba40ba04362fe37130ea0f930/lib/pronto/standardrb.rb#L3

which will also load (and thus, run) it and lead to conflicting outputs since the base rubocop configuration is quite different from `standard`. 

so I could basically copy-paste everything from `pronto-rubocop` and not require it at all, but being all tuned to DRY implementations this really makes me want to tear my hair out...

---

TL;DR: Runners cannot be easily subclassed.

ashkulz:
@julianrubisch can you check if the linked PR (_which is now merged_) solves your problem?

julianrubisch:
does look good!

PR #371, opened 2020-02-25
Allow to filter runners via config

povilasjurcys:
This feature allows to specify which runners will be triggered during `pronto run`. Might solve #365.

# Reason

I use pronto:
* on local machine
* on CI
* as git pre-commit hook
* as git pre-push hook

In each of those cases I would like to run different linters. For example I want to run all pronto linters on CI and only fast-enough linters on git pre-commit hook. With this feature I could add `PRONTO_SKIP_RUNNERS` or `PRONTO_RUNNERS` env variable in order to control which runners will be triggered.

I know that there is a `-r` (aka `--runner`) flag, but:
* it does not have `--skip-runner` options;
* it does not work very well, especially with other flags (some times it tries to load file `pronto/-r` file);
* it does not skip runners which where already loaded.

# Filtering

## Whiteliting runners via config

```yaml
# .pronto.yml
runners:
  - rubocop
  - reek
```

This will ensure that only `rubocop` and `reek` runners will run.

## Blacklisting runners via config

also it's possible to blacklist runners (filter out) like this:
```yaml
# .pronto.yml
skip_runners:
  - rubocop
  - reek
```

## Filtering via env variable

Filtering can be done with env variables:
```bash
# whitelist:
PRONTO_RUNNERS="rubocop reek" pronto run

# blacklist:
PRONTO_SKIP_RUNNERS="rubocop reek" pronto run
```

# Other notable things

I could not resist to do small refactoring in `config.rb` and `runners.rb` classes, I hope it's not a big deal. Let me know if I need to open separate ticket just for refactoring.

Also I'm not sure if config structure is good enough. As an alternative I see nested structure for runners like this:

```yaml
runners:
  only:
    - reek
    - rubocop
  except:
    - rubocop
```

Would be nice to receive some feedback about it.


povilasjurcys:
@doomspork can you check my PR?

.gitignore
```diff
@@ -4,3 +4,5 @@ pkg/*
 .DS_Store
 Gemfile.lock
 coverage
+.history/
```
mknapik: I'd leave out editor specific configuration.
Personally, I'd recommend configuring [global gitignore](https://stackoverflow.com/a/7335487).

lib/pronto/config.rb
```diff
@@ -52,27 +52,57 @@ def bitbucket_hostname
...
-        verbose = ENV['PRONTO_VERBOSE'] || @config_hash['verbose']
+        verbose = fetch_value('verbose')
         verbose ? Logger.new($stdout) : Logger.silent
       end
     end
+
+    private
+
+    def fetch_integer(key)
+      env_key = env_key(key)
```
mknapik: Could you rename a variable to avoid shadowing a method?

lib/pronto/runners.rb
```diff
@@ -6,25 +6,42 @@ def initialize(runners = Runner.runners, config = Config.new)
...
     private
 
+    attr_reader :config, :runners
+
+    def active_runners
+      runners.select { |runner| active_runner?(runner) }
+    end
+
+    def active_runner?(runner)
+      return true if config.runners.empty? && config.skip_runners.empty?
```
mknapik: :+1: for keeping it backward compatible

lib/pronto/config.rb
```diff
@@ -52,27 +52,57 @@ def bitbucket_hostname
...
+      ENV[env_key(key)] || @config_hash[key]
+    end
+
+    def env_key(key)
+      "PRONTO_#{key.upcase}"
+    end
+
+    def fetch_list(key)
+      Array(fetch_value(key)).flat_map do |runners|
+        runners.split(' ')
```
mknapik: I'm not sure what is the most common separator for CLI tools.
How about using `,` instead?

Then quotes can be omitted in the shell command:

```bash
PRONTO_RUNNERS=rubocop,reek pronto run
```

.gitignore
povilasjurcys: thanks, you just saved me from some repetitive work :heart_eyes:

lib/pronto/config.rb
povilasjurcys: changed to `,`

povilasjurcys:
> Personally, I like a nested config more.

@mknapik are you talking about this case:
```yaml
# .pronto.yml
skip_runners:
  - rubocop
  - reek
```

or something completely different?

> I'd love to see tests with a `runners` context with clear 4 cases:
> 
> * no `only` nor `except`
> * just `only`
> * just `except`
> * both `only` and `except`

Added more tests. Let me know if there is something else missing.

mknapik:
You've proposed a structure like:

> Also I'm not sure if config structure is good enough. As an alternative I see nested structure for runners like this:
>
>runners:
>  only:
>    - reek
>    - rubocop
>  except:
>    - rubocop

and I think I like it more.
It would also probably require adjusting env vars accordingly (probably `PRONTO_RUNNERS_EXCEPT` and `PRONTO_RUNNERS_ONLY`).
@prontolabs/core Does anyone has on opinion on config format?

.gitignore
```diff
@@ -4,5 +4,4 @@ pkg/*
 .DS_Store
 Gemfile.lock
 coverage
-.history/
 .vscode/
```
mknapik: You've missed `.vscode` (you could amend the commit if you want to)

spec/pronto/config_spec.rb
```diff
@@ -133,14 +133,33 @@ module Pronto
     describe '#skip_runners' do
       subject { config.skip_runners }
 
-      context 'when there is an entry in the config file' do
+      let(:env_variables) { {} }
```
mknapik: :+1:

.gitignore
povilasjurcys: sorry about that. Fixed and squashed commits

ashkulz:
@povilasjurcys can you rebase this PR?

povilasjurcys:
@ashkulz rebased the code :+1:

ashkulz:
Hmm, not sure why the CI ran twice and failed for `pull_request` -- do you have any idea, @dsander?

dsander:
@ashkulz Not super sure to be honest, maybe #398 fixes that. Generally I think it was wrong to use `pull_request_target` in the action.

ashkulz:
@povilasjurcys can I trouble you to merge/rebase again? GitHub CI failed for 2.4, hopefully #398 fixed that 🤞

ashkulz:
Also, I'd appreciate a change to `README.md` which documents this change as well 🙂

povilasjurcys:
> Also, I'd appreciate a change to `README.md` which documents this change as well

@ashkulz I added [two lines](https://github.com/prontolabs/pronto/pull/371/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R293-R294) in the readme. Let me know if that's enough or not.

> can I trouble you to merge/rebase again? GitHub CI failed for 2.4, hopefully #398 fixed that crossed_fingers

Sadly tests for ruby 2.4 is still failing. I highly doubt that it's because of changes in this PR

ashkulz:
> I added [two lines](https://github.com/prontolabs/pronto/pull/371/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R293-R294) in the readme. Let me know if that's enough or not.

I think you need to elaborate what happens by default i.e. both the options are not provided, otherwise looks good to me!

> Sadly tests for ruby 2.4 is still failing. I highly doubt that it's because of changes in this PR

Sorry to trouble you -- can you merge `master` again? @dsander made a PR which should fix that.

povilasjurcys:
> I think you need to elaborate what happens by default i.e. both the options are not provided, otherwise looks good to me!

Added separate table with some top-level properties [here](https://github.com/prontolabs/pronto/pull/371/commits/a060c2e7b1b0f6b74daa2ca2b58f6a915184a719). Since it's a new thing, I would like to hear your opinion

povilasjurcys:
@ashkulz can we merge this PR?

jarl-dk:
@ashkulz: It seems like this one can be merged now, right?

ashkulz:
Thanks for the contribution, @povilasjurcys! Sorry for the delay in reviewing :see_no_evil:

PR #379, opened 2020-05-13
Relax rugged gem dependency

renatolond:
Rugged just released v1.0 to match libgit. It is expected to be API compatible with 0.28, which is already supported by pronto. (from https://github.com/libgit2/libgit2/releases)

This relaxes the rugged dependency to accept 1.0.x releases.


renatolond:
I'll check when I have some time, maybe it's already a change in one of the 0.20 series that was not caught up before.

renatolond:
There, I've tested locally with 0.28 and 0.99 and it seems that #380 fixes the tests that are broken here

PR #380, opened 2020-05-13
Fix commits until walk

renatolond:
It seems rugged < 0.28 also accepts "Commits" objects, but starting 0.99
it will only accept SHAs for the walk. It seems to work for versions
before 0.28 as well.

### DIFF
--- a/lib/pronto/git/repository.rb
+++ b/lib/pronto/git/repository.rb
@@ -38,7 +38,7 @@ module Pronto
 
       def commits_until(sha)
         result = []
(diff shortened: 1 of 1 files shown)


renatolond:
I've tested this against 0.28, 0.99 and 1.0, not sure if it's worth to test more in the past?

mknapik:
It seems `rugged` is quite fragile in terms of compiling on various ruby/libgit2 versions so it's quite difficult to test all possibilities.
IMO supporting those versions would be good enough.
We could change required version in the gemspec. @prontolabs/core What do you think about requiring version '~> 0.28'?

PR #389, opened 2020-11-20
Readme: PR number can be taken from github context without parsing

Vasfed:
In github actions there's no need to parse the payload file with `jq`, PR number is already available in `github.event.pull_request.number`
Also more secure and readable way is to pass env variables via `env`


ashkulz (approved):
Thanks, @Vasfed!

PR #370, opened 2020-01-08
Update thor dependency

rwojnarowski:
Thor changelog:
https://github.com/erikhuda/thor/blob/master/CHANGELOG.md

PR #393, opened 2020-12-24
Updating thor dependency version

jer0m:
Upgrade thor dependency to v 1.0 minimum to fit Rails 6.1 needs.


ashkulz:
Can you explain why you need to bump the minimum version up? Maybe you just need to include thor in the `bundle update` command.

jer0m:
Well, apart the Rails 6.1 upgrade, I believe it's a good idea to keep dependencies up-to-date.

ashkulz:
For a tool like pronto, supporting the lowest version possible is better as then there won't be any conflicts preventing its usage in projects. The dependency is also `>=`, so it's unlikely to be the cause of a conflict (as compared to `~>`).

ashkulz:
Okay, so I visited the linked issue -- #370 already did what you want, so I'm closing this.

jer0m:
Ok didn't see the other PR, I understand for the dependency matching. Thanks

PR #397, opened 2021-01-30
 Run CI on GitHub Actions

dsander:
Ruby 2.3 has some strange openssl version mismatch issue on GitHub
Actions. The `rugged` gem itself installs fine but when it loads it's C
extensions it somehow loads the wrong openssl version. Running the specs
interactively (even on GitHub Actions) works without issues.

Sorry for the commit mess, getting this to work(ish) was harder than expected. I think the Actions are not run in the PR because I am not a maintainer, The output can be checked here: https://github.com/dsander/pronto/actions/runs/523525169

Sadly I have no clue how we can get the tests on Ruby 2.3 to work. 

Rubocop comments are working: https://github.com/dsander/pronto/pull/1

Labeled as draft because the switch might mean to drop CI for Ruby 2.3.


pronto.gemspec
```diff
@@ -54,5 +54,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec-its', '~> 1.2')
   s.add_development_dependency('rspec-expectations', '~> 3.4')
   s.add_development_dependency('rubocop', '~> 0.58')
-  s.add_development_dependency('simplecov', '~> 0.14')
+  s.add_development_dependency('simplecov', '~> 0.14', '!= 0.18.0', '!= 0.18.1', '!= 0.18.2', '!= 0.18.3', '!= 0.18.4',
+                               '!= 0.18.5', '!= 0.19.0', '!= 0.19.1')
```
ashkulz, line 58: Can you explain this, preferably with a comment?

ashkulz (commented):
Did you try running without `bundle exec`? Also, what's the version of bundler on the Ruby 2.3 run -- maybe you specifically need to use a version to enable support for such an old version?

ashkulz:
@dsander I'm open to dropping Ruby 2.3 support -- I think Ruby 2.5 is going to be EOL in a few months, so makes sense to drop something which can't run in CI. What do you think?

dsander:
> 
> 
> Did you try running without `bundle exec`? Also, what's the version of bundler on the Ruby 2.3 run -- maybe you specifically need to use a version to enable support for such an old version?

I don't think bundler is at fault here, the gems load without issues. I think what happens is that `rugged` builds it's bundled `libgit` against the correct - bundled with ruby 2.3 - openssl 1.0 library. Something,, for some reason then loads system installed openssl 1.1 library at runtime and segfaults. This issue seems to describe the same problem https://github.com/libgit2/rugged/issues/718.

I don't even know where to point the finger, is it the github action stack, the `ruby/setup-ruby` action, rugged, the way rugged compiles libgit, how ruby builds C extensions or how ruby/C loads libraries?

What makes things worse, I have no clue is why everything works normally when the specs are run interactively "inside" the action when connecting to it via ssh.

> @dsander I'm open to dropping Ruby 2.3 support -- I think Ruby 2.5 is going to be EOL in a few months, so makes sense to drop something which can't run in CI. What do you think? I don't want to keep Travis CI just for the 2.3 support.

More than happy to drop it, 2.3 is EOL for almost two years, so hopefully not many people are still using it 😄

pronto.gemspec
dsander: Sure, done.

dsander:
If we drop Ruby 2.3 from CI, the rubocop version is also bumped (or we need to enforce the older version in the gemspec. It not complains about new cops not being enabled: https://github.com/dsander/pronto/runs/1800709555?check_suite_focus=true#step:5:11

I don't think enabling all new cops is an option because everyone runs a different rubocop version locally

pronto.gemspec
```diff
@@ -54,5 +54,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec-its', '~> 1.2')
   s.add_development_dependency('rspec-expectations', '~> 3.4')
   s.add_development_dependency('rubocop', '~> 0.58')
-  s.add_development_dependency('simplecov', '~> 0.14')
+  # Enforces a rubocop version that is supported by CodeClimate, they support 0.17.x or >= 0.20
```
ashkulz, line 57: Can you add a link on where this is documented? Also, your comment mentions `rubocop` but the dependency is for `simplecov`.

.github/workflows/action.yml
```diff
@@ -0,0 +1,41 @@
...
+          - "3.0"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby-version }}
+          bundler-cache: true
+      - name: Test & publish code coverage
+        uses: paambaati/codeclimate-action@v2.7.4
```
ashkulz, line 29: Is there an advantage of using this action, rather than simply using `curl` to download the binary like the TravisCI script does?

ashkulz:
I think we should leave Ruby 2.3 in, but [explicitly exclude it like done for pronto-haml](https://github.com/prontolabs/pronto-haml/blob/8e230f3/.github/workflows/checks.yml#L15), so that it's present in the definition and is documented _why_ it fails but doesn't actually get run -- using `continue-on-error` isn't great, see actions/toolkit#399

.github/workflows/action.yml
dsander: I didn't read every line of the action super carefully (and I am not familiar with the cc test reporter), but I think it mainly sets up [required environment varibales](https://github.com/paambaati/codeclimate-action/blob/0f8af43fca84b500025ca48b581bcff933244252/src/main.ts#L39-L53) and handles some edge cases.

pronto.gemspec
dsander: Sorry about that, fixed.

dsander:
> I think we should leave Ruby 2.3 in, but [explicitly exclude it like done for pronto-haml](https://github.com/prontolabs/pronto-haml/blob/8e230f3/.github/workflows/checks.yml#L15), so that it's present in the definition and is documented _why_ it fails but doesn't actually get run

Sure, that works as well. I thought you were implying dropping Ruby 2.3 support all together.

.github/workflows/action.yml
ashkulz: I'd prefer to use the `curl` approach as that's the official route, rather than depending on a third-party action which may need to be reviewed manually or bumped via dependabot. If you **do** want to use it, please use the commit hash of latest release (`v2.7.5`) -- that's a known good point :slightly_smiling_face: .

pronto.gemspec
```diff
@@ -54,5 +54,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec-its', '~> 1.2')
   s.add_development_dependency('rspec-expectations', '~> 3.4')
   s.add_development_dependency('rubocop', '~> 0.58')
-  s.add_development_dependency('simplecov', '~> 0.14')
+  # Enforces a simplecov version that is supported by CodeClimate, they support 0.17.x or >= 0.20 (https://docs.codeclimate.com/docs/configuring-test-coverage)
+  s.add_development_dependency('simplecov', '~> 0.17', '!= 0.18.0', '!= 0.18.1', '!= 0.18.2', '!= 0.18.3', '!= 0.18.4',
+                               '!= 0.18.5', '!= 0.19.0', '!= 0.19.1')
```
ashkulz, line 59: Considering it's a development-only dependency, I think it's just simpler to do
```suggestion
  s.add_development_dependency('simplecov', '~> 0.20') # see https://docs.codeclimate.com/docs/configuring-test-coverage
```
What do you think?

.github/workflows/action.yml
```diff
@@ -0,0 +1,44 @@
...
+      matrix:
+        ruby-version:
+          - "2.3"
+          - "2.4"
+          - "2.5"
+          - "2.6"
+          - "2.7"
+          - "3.0"
+        exclude:
+          - ruby: "2.3" # Rugged uses the wrong openssl version on CI and segfaults (similar to https://github.com/libgit2/rugged/issues/718)
```
ashkulz, line 23: Appreciate the detailed comment :+1:

ashkulz (commented):
Thanks for your patience, @dsander! Sorry if I wasn't clear above, I definitely didn't mean to drop Ruby 2.3 support immediately -- that would mean the next release would've to be `0.12.0` (as it's a breaking change).

ashkulz:
Can you also remove `.travis.yml` as part of this PR? I think we're good to go, otherwise :+1:

.github/workflows/action.yml
dsander: I locked the version of the action to the commit of the latest release, this should make it a bit safer and still use the same code even if the version is re-tagged. It looks like the action is semi-official, codeclimate reference it in their documentation: https://docs.codeclimate.com/docs/github-actions-test-coverage

.github/workflows/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        ruby-version:
+          - "2.3"
+          - "2.4"
+          - "2.5"
+          - "2.6"
+          - "2.7"
+          - "3.0"
```
ashkulz, line 21: NIT: You can do it in one-line
```suggestion
        ruby-version: ['2.3', '2.4', '2.5', '2.6', '2.7', '3.0']
```

pronto.gemspec
```diff
@@ -54,5 +54,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency('rspec-its', '~> 1.2')
   s.add_development_dependency('rspec-expectations', '~> 3.4')
   s.add_development_dependency('rubocop', '~> 0.58')
-  s.add_development_dependency('simplecov', '~> 0.14')
+  s.add_development_dependency('simplecov', '~> 0.17', '!= 0.18.0', '!= 0.18.1', '!= 0.18.2', '!= 0.18.3', '!= 0.18.4',
+                               '!= 0.18.5', '!= 0.19.0', '!= 0.19.1') # see https://docs.codeclimate.com/docs/configuring-test-coverage
```
ashkulz, line 58: Like I said earlier, why don't we just bump it to `~> 0.20.0`?

ashkulz (approved):

ashkulz:
Thanks, @dsander! Sorry about not merging this earlier 🙁

pronto.gemspec
dsander: Sorry, I missed the notification. `0.20.0` doesn't support ruby 2.4 anymore.

PR #398, opened 2021-02-19
Fix duplicated GitHub action runs

dsander:
It is not recommended to use `pull_request_target` when code is executed
that could modify the repository.

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -5,7 +5,6 @@ on:
     branches: [master]
   pull_request:
     branches: [master]
-  pull_request_target:
 
 jobs:
   test:


dsander:
When this is merged pronto will probably not be able to comment on PR anymore because it does not have the right permissions. This is **really** difficult to test because the actions seem to behave differently depending on the repository they are running from/on. The same PR on my fork passes CI: https://github.com/dsander/pronto/pull/2

I think what we want is something like the setup described [here](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) (search for `CommentPR.yml`). Do you know how the commenting worked with TravisCI? Was there a GitHub token in the CI secrets?

ashkulz:
It's fine if we remove `pull_request_target` -- but why is the `pull_request` failing for Ruby 2.4? It happens on this PR as well.

ashkulz:
I created #399 which still ran the duplicated runs, but weirdly enough it was successful for 2.4 🤷‍♂️

ashkulz (approved):

dsander:
I think that is/was because depending on the trigger and origin of the PR (either a fork or on this repository) the commits are checked out differently, but I am not 100% sure.

PR #403, opened 2021-03-05
 Run pronto on GitHub Actions without any reporting 

dsander:
In #401 we discussed that there
currently is no 100% safe way to run pronto on `pull_request_target`
actions we can not send any reports to the GitHub pull request.

This simple action will show up as a failed check. The offences then
need to be viewed in the action output.

### DIFF
--- a/.github/workflows/pronto.yml
+++ b/.github/workflows/pronto.yml
@@ -1,0 +1,20 @@
+name: Pronto
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  pronto:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.5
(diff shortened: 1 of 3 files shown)


dsander:
The test commit that included a rubocop offense failed the action as expected: https://github.com/prontolabs/pronto/runs/2041862031?check_suite_focus=true#step:4:228

dsander:
I removed 2.4 from the test matrix because of this failures https://github.com/prontolabs/pronto/runs/2041832382?check_suite_focus=true#step:3:33

ashkulz:
Makes sense about the GitHub CI check for now! For the Ruby 2.4 failure, can't we [override the gem version for 2.4 like it's done here](https://github.com/prontolabs/pronto-rubocop/blob/master/.github/workflows/checks.yml#L26)? We'll have to add [support for Gemfile.local in Gemfile](https://github.com/prontolabs/pronto-rubocop/blob/master/Gemfile#L7) but I think that's fine.

dsander:
We can, I simply wasn't aware of that trick 😄

ashkulz (approved):
Thanks, @dsander!

PR #408, opened 2021-05-05
use rexml => 3.2.5 due to CVE-2021-28965

swiknaba:
https://github.com/advisories/GHSA-8cr8-4vfw-mr7h

> The REXML gem before 3.2.5 in Ruby before 2.6.7, 2.7.x before 2.7.3, and 3.x before 3.0.1 does not properly address XML round-trip issues. An incorrect document can be produced after parsing and serializing.


ashkulz (approved):
Thanks for the contribution!

PR #410, opened 2021-07-08
Allow the default commit to run against to be configured

ozydingo:
## Use Case

For repositories whose mainline branch is not `master` (new Github projects now default to `main`, and some projects require pull requests against `development` or similar), allow the default target branch for pronto to be configured. If not configured, it remains `master`.

## Notes

Similar to other congiruations, it can be specified in `.pronto.yml` (using `default_commit`) or by environment variable (`PRONTO_DEFAULT_COMMIT`).

The `-c` flag is still supported to specify a target commit. Nevertheless, configuring a default makes sense when the existing default `master` does not even exist in a project as is the case for many new Github repos.

Issue: https://github.com/prontolabs/pronto/issues/412

### DIFF
--- a/lib/pronto.rb
+++ b/lib/pronto.rb
@@ -55,9 +55,9 @@ require 'pronto/formatter/null_formatter'
 require 'pronto/formatter/formatter'
 
 module Pronto
-  def self.run(commit = 'master', repo_path = '.',
+  def self.run(commit = nil, repo_path = '.',
                formatters = [Formatter::TextFormatter.new], file = nil)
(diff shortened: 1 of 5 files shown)


ozydingo:
@mknapik  you've looked at some other config-related PRs, any time to give this one a quick glance? 🙏

ashkulz (approved):

ashkulz:
Thanks for the contribution, @ozydingo! Sorry for the delay in reviewing :see_no_evil:

PR #415, opened 2021-11-03
Remove deprecated runner on README

MatheusPoliCamilo:
Removed `pronto-bundler_audit` from runners list on Readme. It is a deprecated runner (as said here: https://github.com/pdobb/pronto-bundler_audit#maintainer-needed) that isn't working.


ashkulz (approved):

ashkulz:
Thanks for the contribution, @MatheusPoliCamilo! In case @pdobb or any maintainer does make it compatible, I'd be fine with adding it back.

PR #421, opened 2022-02-20
Removing old bitbucket pull request commits

### LINKED ISSUE #255: Delete fixed comments
One thing that happens on big pull requests is that there are many old comments made by pronto.

image

Would it be possible to add an option to delete old comments made by the bot before new ones are made? This means existing issues are reported again but fixed issues are gone.

misteral:
Fix #255 for Bitbucket Pull Request format.

### DIFF
--- a/lib/pronto/formatter/bitbucket_pull_request_formatter.rb
+++ b/lib/pronto/formatter/bitbucket_pull_request_formatter.rb
@@ -1,6 +1,27 @@
 module Pronto
   module Formatter
     class BitbucketPullRequestFormatter < PullRequestFormatter
+      def format(messages, repo, patches)
+        client = client_module.new(repo)
+        existing = existing_comments(messages, client, repo)
+        comments = new_comments(messages, patches)
+        additions = remove_duplicate_comments(existing, comments)
+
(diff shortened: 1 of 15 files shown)


lib/pronto/formatter/bitbucket_pull_request_formatter.rb
```diff
@@ -1,6 +1,27 @@
 module Pronto
   module Formatter
     class BitbucketPullRequestFormatter < PullRequestFormatter
+      def format(messages, repo, patches)
+        client = client_module.new(repo)
+        existing = existing_comments(messages, client, repo)
+        comments = new_comments(messages, patches)
+        additions = remove_duplicate_comments(existing, comments)
+
+        remove_comments(client, existing, comments) if existing.any?
```
ashkulz, line 10: This should be optional. I'm not sure how we'd go about it, maybe a separate `PullRequestFormatter` which deletes comments?

ashkulz (changes requested):
Sorry for the very late delay. Would appreciate merging with `master` to ensure we pick up the latest CI changes :+1:

ashkulz:
Closing since no update for 2+ years.

PR #423, opened 2022-04-01
Makes message for git formatters to prevent confusion

pftg:
When Developer see that there are 0 messages posted in the log, but pronto still is failed, there is no idea what's going on.

Added more insights in the message in order to prevent such confusion.

### DIFF
--- a/lib/pronto/formatter/git_formatter.rb
+++ b/lib/pronto/formatter/git_formatter.rb
@@ -10,7 +10,7 @@ module Pronto
         
         approve_pull_request(comments.count, additions.count, client) if defined?(self.approve_pull_request)
 
-        "#{additions.count} Pronto messages posted to #{pretty_name}"
+        "#{additions.count} Pronto messages posted to #{pretty_name} (#{existing.count} existing)"
       end
 
       def client_module
--- a/spec/pronto/formatter/github_formatter_spec.rb
+++ b/spec/pronto/formatter/github_formatter_spec.rb
@@ -7,9 +7,11 @@ module Pronto
         subject { formatter.format(messages, repo, nil) }
 
         let(:repo) { Git::Repository.new('spec/fixtures/test.git') }
-        let(:published_comments_msg) { "%{count} Pronto messages posted to #{formatter.pretty_name}" }
+        let(:published_comments_msg) do
+          "%<count>i Pronto messages posted to #{formatter.pretty_name} (%<existing_count>i existing)"
+        end
 
-        context 'with duplicates' do
+        context 'with duplicates in the new messages' do
           let(:messages) { [message, message] }
           let(:message) { Message.new('path/to', line, :warning, 'crucial') }
           let(:line) { double(new_lineno: 1, commit_sha: '123', position: nil) }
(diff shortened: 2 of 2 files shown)


lib/pronto/formatter/git_formatter.rb
```diff
@@ -10,7 +10,7 @@ def format(messages, repo, patches)
         
         approve_pull_request(comments.count, additions.count, client) if defined?(self.approve_pull_request)
 
-        "#{additions.count} Pronto messages posted to #{pretty_name}"
+        "New #{additions.count} (to #{comments.count} existed) Pronto messages posted to #{pretty_name}"
```
ashkulz, line 13:
```suggestion
        "#{additions.count} Pronto messages posted to #{pretty_name} (#{comments.count} existing)"
```
I'd prefer the additional information was at the end.

ashkulz (changes requested):
Thanks for the contribution, @pftg!

pftg:
Hey @ashkulz updated PR with your requested changes

spec/pronto/formatter/github_formatter_spec.rb
```diff
@@ -7,7 +7,9 @@ module Formatter
         subject { formatter.format(messages, repo, nil) }
 
         let(:repo) { Git::Repository.new('spec/fixtures/test.git') }
-        let(:published_comments_msg) { "%{count} Pronto messages posted to #{formatter.pretty_name}" }
+        let(:published_comments_msg) do
+          "%<new_count>i Pronto messages posted to #{formatter.pretty_name} (%<existed_count>i existing)"
```
ashkulz, line 11:
```suggestion
          "%<count>i Pronto messages posted to #{formatter.pretty_name} (%<existing_count>i existing)"
```
**nit**: better variable names

ashkulz (approved):

spec/pronto/formatter/github_formatter_spec.rb
pftg: 👍

pftg:
@ashkulz added suggested changes to the pr

spec/pronto/formatter/github_formatter_spec.rb
```diff
@@ -25,7 +27,7 @@ module Formatter
               .should_receive(:create_commit_comment)
               .once
 
-            subject.should eq format(published_comments_msg, count: 1)
+            subject.should eq format(published_comments_msg, count: 1, existing_count: 1)
```
ashkulz, line 30: I'm not sure this makes sense, as it means that 1 comments added, and 1 existing. I'd rather expect 0 existing. Maybe we need to use `comments.count - additions.count` instead?

pftg: Originally I tried to use `existing` comments, will re-check once more, maybe I missed something

pftg:
@ashkulz updated please recheck

ashkulz (approved):
Awesome, @pftg!

ashkulz:
Thanks for the contribution!

PR #426, opened 2022-04-19
Fix: line pull request formatter

abr2001:


ashkulz:
@abr2001 can you add more details on why this was required, preferably with a test or a description how to trigger the scenario you're trying to fix?

ashkulz:
Any update, @abr2001?

Issue #430, opened 2022-07-12
Relax Octokit version dependency


pbstriker38:
Octokit v5.0.0 was released. The only change is the drop of support for EOL Ruby's.

https://github.com/octokit/octokit.rb/releases/tag/v5.0.0

ashkulz:
Release `0.11.1` has been pushed to RubyGems.

PR #433, opened 2022-09-15
Relax octokit version dependency

andriusm:
Previous constrain was limiting where Danger 9.x cannot be used together with Pronto in a project because of different octokit dependency version requirements. There was also an issue registered suggesting this - https://github.com/prontolabs/pronto/issues/430

### DIFF
--- a/pronto.gemspec
+++ b/pronto.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   EOF
 
   s.licenses = ['MIT']
-  s.required_ruby_version = '>= 2.3.0'
+  s.required_ruby_version = '>= 2.7.0'
   s.rubygems_version = '1.8.23'
 
   s.files = `git ls-files`.split($RS).reject do |file|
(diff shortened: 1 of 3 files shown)


avelicka (approved):

avelicka (approved):

lib/pronto/version.rb
```diff
@@ -1,6 +1,6 @@
 module Pronto
   module Version
-    STRING = '0.11.0'.freeze
```
ashkulz, line 3: Please don't change this.

pronto.gemspec
```diff
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency('gitlab', '~> 4.4', '>= 4.4.0')
   s.add_runtime_dependency('httparty', '>= 0.13.7')
-  s.add_runtime_dependency('octokit', '~> 4.7', '>= 4.7.0')
+  s.add_runtime_dependency('octokit', '~> 5.0')
```
ashkulz, line 45:
```suggestion
  s.add_runtime_dependency('octokit', '>= 4.7.0')
```
Maybe just remove the `~> 4.7` part?

ashkulz (changes requested):

andriusm:
Changed octokit dependency to be `>= 4.7.0` and removed several versions of Ruby from the build matrix which were failing. Octokit version 5.0 dropped support for Ruby versions 2.0, 2.1, 2.2, 2.3, 2.4, 2.5 and 2.6, which have all reached end-of-life.

tomasv (approved):

ashkulz (changes requested):
I wouldn't like to drop support for Ruby 2.6 although it's EOL, because that's the currently supported JRuby version.

Dropping support for old Ruby versions will also be problematic, as that'll require a major version bump. Is there a way to fix the CI without doing all this?

ashkulz:
Any update, @andriusm?

PR #435, opened 2022-10-21
Support specifying the config file path in an environment variable

hosamaly:
This is a simple solution to #226 that introduces support for the environment variable `PRONTO_CONFIG_FILE`.

There was a previous attempt at solving this issue in #228, but it's much more elaborate. This is a simpler solution because it doesn't try to add a CLI option.


ashkulz:
Thanks for the contribution, @hosamaly!

PR #450, opened 2023-10-22
Update how formatters are registered, allowing the addition of custom formatter

emilio2hd:
Use the same Devise logic to registry the formatters, instead of using a frozen const.

Please check https://github.com/prontolabs/pronto/issues/451 for more context

### DIFF
--- a/lib/pronto/formatter/formatter.rb
+++ b/lib/pronto/formatter/formatter.rb
@@ -1,30 +1,32 @@
 module Pronto
   module Formatter
-    def self.get(names)
-      names ||= 'text'
-      Array(names).map { |name| FORMATTERS[name.to_s] || TextFormatter }
-        .uniq.map(&:new)
-    end
+    class << self
+      def add(label, formatter_klass = nil)
+        unless formatter_klass.method_defined?(:format)
+          raise NoMethodError, "format method is not declared in the #{label} class."
+        end
 
-    def self.names
-      FORMATTERS.keys
-    end
+        base = Pronto::Formatter::Base
+        raise "#{label.inspect} is not a #{base}" unless formatter_klass.ancestors.include?(base)
 
-    FORMATTERS = {
-      'github' => GithubFormatter,
-      'github_status' => GithubStatusFormatter,
-      'github_combined_status' => GithubCombinedStatusFormatter,
-      'github_pr' => GithubPullRequestFormatter,
-      'github_pr_review' => GithubPullRequestReviewFormatter,
-      'gitlab' => GitlabFormatter,
-      'gitlab_mr' => GitlabMergeRequestReviewFormatter,
-      'bitbucket' => BitbucketFormatter,
-      'bitbucket_pr' => BitbucketPullRequestFormatter,
-      'bitbucket_server_pr' => BitbucketServerPullRequestFormatter,
-      'json' => JsonFormatter,
-      'checkstyle' => CheckstyleFormatter,
-      'text' => TextFormatter,
-      'null' => NullFormatter
-    }.freeze
+        raise ArgumentError, "formatter #{label.inspect} has already been added." if formatters.key?(label)
+
+        formatters[label] = formatter_klass
+      end
+
+      def get(names)
+        names ||= 'text'
+        Array(names).map { |name| formatters[name.to_s] || TextFormatter }
+          .uniq.map(&:new)
+      end
+
+      def names
+        formatters.keys
+      end
+
+      def formatters
+        @formatters ||= {}
+      end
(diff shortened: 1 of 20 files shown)


lib/pronto/formatter/formatter.rb
```diff
@@ -1,30 +1,32 @@
...
+        end
 
-    def self.names
-      FORMATTERS.keys
-    end
+        base = Pronto::Formatter::Base
+        raise "#{label.inspect} is not a #{base}" unless formatter_klass.ancestors.include?(base)
+
+        raise ArgumentError, "formatter #{label.inspect} has already been added." if formatters.key?(label)
+
```
ashkulz, line 13: **concern**: I'm not so sure about this part, what happens if the code gets reloaded or required in some way which causes the same formatter to be added twice? Normally, this shouldn't be a concern but I'm a bit wary about guessing about the way people use Pronto.

lib/pronto/formatter/formatter.rb
```diff
@@ -1,30 +1,32 @@
 module Pronto
   module Formatter
-    def self.get(names)
-      names ||= 'text'
-      Array(names).map { |name| FORMATTERS[name.to_s] || TextFormatter }
-        .uniq.map(&:new)
-    end
+    class << self
+      def add(label, formatter_klass = nil)
```
ashkulz, line 4:
```suggestion
      def register(label, formatter_klass)
```
**suggestion**: why do you want to make `formatter_klass` optional? It would raise an exception below anyway :man_shrugging:

Also, `register` sounds a much better name as we're not providing `remove` method.

lib/pronto/formatter/formatter.rb
```diff
@@ -1,30 +1,32 @@
...
+          .uniq.map(&:new)
+      end
+
+      def names
+        formatters.keys
+      end
+
+      def formatters
+        @formatters ||= {}
+      end
```
ashkulz, line 29: **thought**: I would prefer not exposing this at all, as the only place this is used is inside `Pronto::Formatter::Base.name`. I'd rather we did away with the `register` method not taking a `name` but overriding it explicitly in every class.

This way, it remains an implementation detail.

ashkulz (changes requested):
Looks **very** good, @emilio2hd! Just a few minor changes and we're good to go.

lib/pronto/formatter/formatter.rb
emilio2hd: Yeah, makes sense.
I haven't thought about reloading the formatter.

lib/pronto/formatter/formatter.rb
emilio2hd: Good catch, thanks!
I forgot to remove `= nil`.
Agreed, `register` is a better name.

lib/pronto/formatter/formatter.rb
emilio2hd: Hmm... Sorry, I'm not sure if I fully understand. Do you mean something like this:

```ruby
module Pronto
  module Formatter
    class TextFormatter < Base
      def option_name
        'text'
      end

      def format(messages, _repo, _patches)
        # formatting stuff...
      end
    end
  end
end

Pronto::Formatter.register(Pronto::Formatter::TextFormatter)
```
Regarding hidding the `Pronto::Formatter::formatters`. Yeah... That's going to require a bit more work.
I've tried to get rid of `.invert[self]` and almost all the tests started to fail and I didn't understand why, so I left it for another time. 😂

emilio2hd: Wait! I think I've got it. You're talking like this:
```ruby
module Formatter
  class Base
    def self.name
      raise NoMethodError, "Must be implemented in subclasses."
    end
    # rest of implementation...
  end
end

# in subclasses
module Formatter
  class TextFormatter < Base
    def self.name
      'text'
    end
    # formatting...
  end
end
```

In this case, there's no need for `Formatter::formatters.invert[self]` anymore!

ashkulz: I think you got me the second time around, but this is still exposed and not yet removed?

ashkulz (approved):

lib/pronto/formatter/formatter.rb
emilio2hd: Sorry, my bad!

ashkulz (approved):

ashkulz:
Thanks for the contribution, @emilio2hd!

emilio2hd:
Thank for the guidance and accepting the contribution! 😃

Issue #447, opened 2023-09-05
Rugged 1.7.1 breaks gem


cooljacob204:
Rugged update 1.7.1 which came out yesterday breaks the gem.

```
/lib/rugged/repository.rb:114:in `diff': undefined method `[]' for nil:NilClass (NoMethodError)

        right.diff(left, opts.merge(:reverse => !opts[:reverse]))
                                                     ^^^^^^^^^^
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.11.1/lib/pronto/git/repository.rb:30:in `diff'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.11.1/lib/pronto.rb:64:in `run'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.11.1/lib/pronto/cli.rb:70:in `block in run'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.11.1/lib/pronto/cli.rb:6[8]:in `chdir'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.11.1/lib/pronto/cli.rb:68:in `run'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/thor-1.2.2/lib/thor/command.rb:27:in `run'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/thor-1.2.2/lib/thor/invocation.rb:127:in `invoke_command'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/thor-1.2.2/lib/thor.rb:3[9]2:in `dispatch'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/thor-1.2.2/lib/thor/base.rb:485:in `start'
	from /opt/hostedtoolcache/Ruby/3.1.4/x64/lib/ruby/gems/3.1.0/gems/pronto-0.[11].1/bin/pronto:6:in `<top (required)>'
[...]

PR #452, opened 2023-10-31
Pin rugged to below 1.7

magni-:
A bandaid until `pronto` is updated to support `rugged` 1.7

Fixes https://github.com/prontolabs/pronto/issues/447


ashkulz:
@magni- instead of merging this, I'd prefer to fix it properly. Would appreciate if you could post a reproduction on #447

magni-:
Unfortunately I don't have the time right now. This is just supposed to be a bandaid, not a final fix, but it would help us avoid having broken dependabot PRs for `pronto` in all of our Ruby repos.

ashkulz:
In that case, I think you should use [the workaround](https://github.com/prontolabs/pronto/issues/447#issuecomment-1726463807) for now, as I don't think it makes sense to merge this and undo it a few days later.

PR #454, opened 2023-11-02
relax octokit dependency to allow releases up to 9.0

### LINKED ISSUE #453: Allow Octokit 8.0
Octokit v8.0.0 was released with changes to the way pull request comments are created. They have removed the deprecated position argument and replaced it with `line`

https://github.com/octokit/octokit.rb/releases/tag/v8.0.0
https://github.com/octokit/octokit.rb/pull/1566

https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request

ashkulz:
fixes #453

### DIFF
--- a/pronto.gemspec
+++ b/pronto.gemspec
@@ -42,7 +42,7 @@ Gem::Specification.new do |s|
 
   s.add_runtime_dependency('gitlab', '>= 4.4.0', '< 5.0')
   s.add_runtime_dependency('httparty', '>= 0.13.7', '< 1.0')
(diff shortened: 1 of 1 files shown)


pbstriker38:
You can't allow Octokit 8.0 without fixing the argument being passed to `create_pull_comment`. Pronto is sending position and Octokit is sending that to the API as `line`. So it will either return a 422 if the position does not map to a valid line or put the comment on the wrong line.

I'm working on a fix and I'll have a PR open shortly.

PR #455, opened 2023-11-03
Remove usage of deprecated position argument for GitHub comments

pbstriker38:
- This allows usage of Octokit v8.0.0

Fixes: #453

### DIFF
--- a/lib/pronto/github.rb
+++ b/lib/pronto/github.rb
@@ -10,8 +10,9 @@ module Pronto
     def pull_comments(sha)
       @comment_cache["#{pull_id}/#{sha}"] ||= begin
         client.pull_comments(slug, pull_id).map do |comment|
-          Comment.new(sha, comment.body, comment.path,
-                      comment.position || comment.original_position)
+          Comment.new(
+            sha, comment.body, comment.path, comment.line || comment.original_line
+          )
         end
       end
     rescue Octokit::NotFound => e
@@ -23,7 +24,7 @@ module Pronto
     def commit_comments(sha)
       @comment_cache[sha.to_s] ||= begin
         client.commit_comments(slug, sha).map do |comment|
-          Comment.new(sha, comment.body, comment.path, comment.position)
+          Comment.new(sha, comment.body, comment.path, comment.line)
         end
       end
     end
@@ -37,9 +38,13 @@ module Pronto
     def create_pull_comment(comment)
       if comment.path && comment.position
         @config.logger.log("Creating pull request comment on #{pull_id}")
-        client.create_pull_comment(slug, pull_id, comment.body,
-                                   pull_sha || comment.sha,
-                                   comment.path, comment.position)
+        client.create_pull_comment(
+          # Depending on the Octokit version the 6th argument can be either postion or line. We'll
+          # provide the `line` as this argument and also provide the line in the options argument.
+          # The API uses `line` and ignores position when `line` is provided.
+          slug, pull_id, comment.body, pull_sha || comment.sha,
+          comment.path, comment.position, { line: comment.position }
+        )
       else
         create_commit_comment(comment)
       end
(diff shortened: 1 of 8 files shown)


ashkulz:
@pbstriker38 I'm not sure this works with older versions of `octokit` -- what do you think of bumping the dependency? Without verifying that it works, I'd be a bit hesitant to merge this as-is.

pbstriker38:
@ashkulz I did verify that it works with both versions. On octokit < 8.0 it sends both `position` and `line` params. The API ignores postion and uses line. On Octokit 8.0 the params are merged into the options and thus the `line` is merged.

The other option I can do here is to not use `Octokit#create_pull_comment` method and to just use `Octokit#post` directly like so:

```ruby
options = {
  body: comment.body,
  commit_id: pull_sha || comment.sha,
  path: comment.path,
  line: comment.position
}
client.post("#{Octokit::Repository.path(slug)}/pulls/#{pull_id}/comments", options)
```

pbstriker38:
I did not test earlier Octokit versions though. I can try and test those too.
- 4.7.0
- 5.x.x
- 6.x.x

lib/pronto/github.rb
```diff
@@ -66,12 +71,11 @@ def create_commit_status(status)
     def create_pull_request_review(comments)
       options = {
         event: @config.github_review_type,
-        accept: 'application/vnd.github.v3.diff+json', # https://developer.github.com/v3/pulls/reviews/#create-a-pull-request-review
         comments: comments.map do |comment|
           {
-            path:     comment.path,
-            position: comment.position,
```
ashkulz, line 73: Sorry, should've been more specific. I meant this invocation -- the other seems to have the fallback in place.

pbstriker38: This one does not need a fallback because the client directly accepts the options instead of using positional params.

ashkulz: So, do all versions of `octokit` accept `line` instead of `position` for this API call?

pbstriker38: Yes. They all accept an options hash that is submitted as the POST body

https://github.com/octokit/octokit.rb/blob/v4.7.0/lib/octokit/client/reviews.rb#L96
https://github.com/octokit/octokit.rb/blob/v5.6.1/lib/octokit/client/reviews.rb#L92
https://github.com/octokit/octokit.rb/blob/v6.1.1/lib/octokit/client/reviews.rb#L92
https://github.com/octokit/octokit.rb/blob/v7.2.0/lib/octokit/client/reviews.rb#L92
https://github.com/octokit/octokit.rb/blob/v8.0.0/lib/octokit/client/reviews.rb#L92

ashkulz: Cool, so the client just passes through to the [server endpoint](https://docs.github.com/en/rest/pulls/reviews?apiVersion=2022-11-28#create-a-review-for-a-pull-request) as-is, which supports both `line` and `position`.

However, I'm a bit confused by the description of the endpoint:
> position: The position in the diff where you want to add a review comment. **Note this value is not the same as the line number in the file**. For help finding the position value, read the note below.

Elsewhere, they say:
> Note: To comment on a specific line in a file, you need to first determine the position of that line in the diff. The GitHub REST API offers the application/vnd.github.v3.diff [media type](https://docs.github.com/rest/overview/media-types#commits-commit-comparison-and-pull-requests). To see a pull request diff, add this media type to the Accept header of a call to the [single pull request](https://docs.github.com/rest/pulls/pulls#get-a-pull-request) endpoint.

> The position value equals the number of lines down from the first "@@" hunk header in the file you want to add a comment. The line just below the "@@" line is position 1, the next line is position 2, and so on. The position in the diff continues to increase through lines of whitespace and additional hunks until the beginning of a new file.

So it looks like using `line` may be better in this case anyway, but I'm not sure I ever saw misplaced comments by pronto so @pbstriker38 would like your view on the interpretation as well before merging :+1:

pbstriker38: Yes, that note is a bit confusing but I think it might be a legacy thing from when the pull request comments API did not return line numbers.

I ran this branch on my fork of pronto and the comments were placed on the correct lines. I can point one of my repos at work to use this branch as well.

pbstriker38: ### Here it is running on one of my repos

image

image

image

image




### And then I fixed some offenses and created a new offense

image

image

ashkulz:
Thanks for the contribution and being patient while we figured things out, @pbstriker38!

ashkulz:
[`0.11.3`](https://rubygems.org/gems/pronto/versions/0.11.3) has been released to RubyGems. Publishing future versions should be simpler due to implementation of Trusted Publishing workflow (c7007deaa060395911ba039604655d6d08fae556) :grin:

PR #462, opened 2024-04-29
improve documentation for integrating with GitLab CI

webit-richardkramer:
When implementing pronto in our GitLab CI we ran into an issue when having more than 50 commits in an MR. In this case, pronto simply showed **all** issues, not only the ones in our changes.

The cause of this was GitLabs git depth (50 by default for us, 20 on newer installations). Unfortunately, it is not enough to separately fetch the target branch (which was already present in the docs), but also to refetch the history of HEAD beyond the initial fetch depth.
You could also do this explicitely, by I found it cleaner to simply disable the git depth for this one job.

In case others stumble upon the same problem, I want to update the docs a bit to reflect this need of disabling git depth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,18 +228,23 @@ On GitLabCI, make sure to run Pronto in a [merge request pipeline](https://docs.
 lint:
   image: ruby:3.3.0 # change to your app's ruby version
   variables:
-    PRONTO_GITLAB_API_ENDPOINT: "https://gitlab.com/api/v4" # required if you are using self hosted Gitlab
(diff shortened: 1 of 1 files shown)


ashkulz (approved):

ashkulz:
Thanks for the contribution, @webit-richardkramer!

ashkulz:
[`0.11.3`](https://rubygems.org/gems/pronto/versions/0.11.3) has been released to RubyGems. Publishing future versions should be simpler due to implementation of Trusted Publishing workflow (c7007deaa060395911ba039604655d6d08fae556) :grin:

PR #464, opened 2024-07-20
Create Integrations section.

MichaelHinrichs:


ashkulz:
I'm not sure we want this -- I don't think too many nested headings make sense.

PR #467, opened 2024-12-12
GitHub - Feature delete outdated comments

datpmt:
- Remove outdated comments that were previously left on the code if the user has pushed new commits addressing those comments.
- Ensure that all resolved comments are cleared to keep the review process clean and up-to-date.

### DIFF
--- a/lib/pronto/comment.rb
+++ b/lib/pronto/comment.rb
@@ -1,5 +1,5 @@
 module Pronto
-  Comment = Struct.new(:sha, :body, :path, :position) do
+  Comment = Struct.new(:sha, :body, :path, :position, :id) do
     def ==(other)
       position == other.position &&
         path == other.path &&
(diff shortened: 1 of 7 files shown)


ashkulz (changes requested):
Sorry for the late feedback, I'm not sure this is a good idea as-is.

* A lot of workflows would prefer that the PR comment is marked as _resolved_, rather than be deleted altogether. It might be acceptable if this is disabled by default, but you can opt-in via some configuration option.
* Sometimes there is conversation on a PR comment from `pronto` and I'm not sure [what the API does](https://docs.github.com/en/rest/pulls/comments#delete-a-review-comment-for-a-pull-request) -- does it just delete the thread or just the comment? Either way, it's not going to look good even if the thread is retained -- would like screenshots of how both scenarios look like.
* It's implemented just for GitHub PRs, but I'm not sure why it couldn't extent to Gitlab MRs which also have the same logic.

ashkulz:
See [this code](https://github.com/Vasfed/pronto-github_resolver/blob/0f44fd8bb069118bf96947ab6f013d10e1de28f4/lib/pronto/github_resolver.rb#L58) for something which should work for resolving the comments, since there's no direct API for this.

ragingdave:
I would say the default should be resolving rather than deleting if possible with a setting to delete as an option. If there was back on forth on a pronto comment thread and it just vanished due to updating a style rule/changing code that artifact of discussion would be lost.

PR #477, opened 2025-04-30
Fix uninitialized constant OpenStruct

krisantuswanandi:
currently running pronto bitbucket_pr will result in error `uninitialized constant BitbucketClient::OpenStruct (NameError)`

the reason it's good up until now (at least in my project) was probably because there are other gems requiring ostruct, for example `rubocop`, through `pronto-rubocop`. after rubocop 1.58.0, it's no longer using ostruct so any pronto (with pronto-rubocop) should fail after upgrading to the latest rubocop.

### DIFF
--- a/lib/pronto/clients/bitbucket_client.rb
+++ b/lib/pronto/clients/bitbucket_client.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'ostruct'
 
 class BitbucketClient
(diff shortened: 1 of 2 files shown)


ashkulz (approved):

ashkulz:
Thanks for the contribution, @krisantuswanandi!

ashkulz:
Now available on RubyGems as [pronto 0.11.4](https://rubygems.org/gems/pronto/versions/0.11.4) :tada:

Issue #478, opened 2025-05-02
Allow Octokit 10.0


pbstriker38:
https://github.com/octokit/octokit.rb/releases/tag/v10.0.0

The methods that I removed for the breaking change are not used by Pronto.

PR #479, opened 2025-05-02
Relax dependency on Octokit to allow v10

pbstriker38:
- No breaking changes in Octokit v10 affect Pronto usage

Fixes: #478


ashkulz (approved):

ashkulz:
Thanks for the contribution, @pbstriker38!

PR #459, opened 2024-01-31
Fix Shallow Checkouts

X-sam:
This addresses Issue #447

PR #487, opened 2025-08-28
Ensure options is always a Hash in diff to avoid Rugged errors

lmonzon4:
## Description

This PR fixes an issue where `Rugged::Repository#diff` could receive `nil` as the `options` argument, causing the following error:

```ruby
undefined method `[]' for nil:NilClass (NoMethodError)
```

This happens because the `diff` method in `Pronto::Git::Repository` passes `options` directly to **Rugged** without ensuring it’s a Hash. When `options` is `nil` (the default value), **Rugged** fails.

## What changed

Initialized `options` with an empty Hash if `nil`:

```ruby
options ||= {}
```

In `:workdir` case simplified `.merge(options || {})` to `.merge(options)` since `options` is guaranteed to be a Hash.

## Why

This makes the method safer and consistent across all cases. Previously:

For `:workdir`, a merge fallback (`options || {}`) was already in place.

For other cases (`:unstaged`, `:index`, `:staged`, `else`), **Rugged** could fail when options was `nil`.

## Example

Before:
```ruby
repo.diff(commit, nil) # => NoMethodError
```

After:
```ruby
repo.diff(commit, {}) # Works as expected
```

No breaking changes. This only ensures `options` is never `nil`.

## Specs
No spec changes needed
image

### DIFF
--- a/lib/pronto/git/repository.rb
+++ b/lib/pronto/git/repository.rb
@@ -8,6 +8,7 @@ module Pronto
       end
 
       def diff(commit, options = nil)
+        options ||= {}
         target, patches = case commit
                           when :unstaged, :index
                             [head_commit_sha, @repo.index.diff(options)]
@@ -22,7 +23,7 @@ module Pronto
                                   include_untracked: true,
                                   include_untracked_content: true,
                                   recurse_untracked_dirs: true
-                                }.merge(options || {})
+                                }.merge(options)
                               )
                             ]
                           else


ashkulz:
@jlurena you said that you were able to get an error
```
undefined method `[]' for nil:NilClass (NoMethodError)
```
`pronto` is mostly used as a CLI, so unless this can be reproduced that way I'm not sure it makes sense to fix a potential bug due to passing a wrong argument to an internal API.

ashkulz:
Ah okay, so looks like you're hitting #447. It is actually an upstream issue (libgit2/rugged#973) and a PR similar yours was closed due to it being an incomplete fix -- see #459.

I'd appreciate if you could help with detailed steps reproducing the original issue -- work stalled as I don't want to merge a workaround without understanding the impact.

Viktor-Ivliev:
look like it fixed in rugged 1.9.4

